PR #17, opened 2021-10-19
All classes are created / All methods are implemented

AntonVolskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
Sofasmile, line 7: let's create `new Lottery()` object before for loop

Sofasmile (approved):

PR #18, opened 2021-10-19
Classes Ball and Lottery were added and methods were implemented.

farfardesu:
Class Ball was added with color and number fields. Also method toStrng was
overriden.
Enum Colors was created.
Method getRandomColor() in class ColorSupplier was implemented.
Lottery class was added with getRandomBall() method.

### DIFF
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,6 +1,6 @@
 package core.basesyntax;
 
-public enum Colors {
+public enum Color {
     RED,
     YELLOW,
     BLACK,
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -5,6 +5,6 @@ import java.util.Random;
 public class ColorSupplier {
     public String getRandomColor() {
         Random random = new Random();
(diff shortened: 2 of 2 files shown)


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = {new Lottery().getRandomBall(),
+                        new Lottery().getRandomBall(),
+                        new Lottery().getRandomBall()};
```
Nick97-git, line 7: let's create a lottery variable one time and call `getRandomBall()` using it. Also, you can move the creation of balls in for loop that you wrote down below but in this case, you will need to use fori loop, not foreach)

Nick97-git (changes requested):

src/main/java/core/basesyntax/Colors.java
```suggestion
public enum Color {
```

Nick97-git (changes requested):

PR #25, opened 2021-10-19
solution is completed

Andrey-Sidorkin:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+        Ball[] balls = new Ball[] {ball1, ball2, ball3};
```
Sofasmile, line 8: let's do it in the loop

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
+        for (Ball theBall : balls) {
```
Sofasmile, line 9:
```suggestion
        for (Ball ball : balls) {
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < 3; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
Sofasmile, line 7: let's create object of `new Lottery()` before the loop and use it in the loop
```suggestion
            balls[i] = lottery.getRandomBall();
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
```
Sofasmile, line 5: better create a constant `NUMBER_OF_BALLS` for number 3 and use it, because now it's a magic number
read more about constants here - https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names  (constant should have `private static final` modifiers)
```suggestion
        Ball[] balls = new Ball[NUMBER_OF_BALLS];
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,10 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+
+    private static int NUMBER_OF_BALLS = 3;
```
Sofasmile, line 5:
```suggestion
    private static final int NUMBER_OF_BALLS = 3;
```

Sofasmile (changes requested):

src/main/java/core/basesyntax/Application.java
Sofasmile: preferable way) 
because when we use `new Lottery()` in for loop we create a new object during each iteration. So when the number of iterations is large, we create a lot of extra objects. We should avoid it

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
 public class Application {
+
```
KseniiaMakarova, line 4: remove this redundant empty line please
we don't need empty lines after class declarations or method signatures

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
KseniiaMakarova, line 10: let's use our constant or array length instead of `3`

Andrey-Sidorkin: I've pushed all the changes, but the button "Re-request review" became unclickable. What is the reason?

KseniiaMakarova (approved):

Sofasmile (approved):

PR #29, opened 2021-10-20
Classes Ball, Lottery were implemented. Enum Color was created.

KostyaNikiforov:
And also I changed a returning data type of method getRandomColor in class ColorSupplier to Color.

### DIFF
--- a/src/main/java/core/basesyntax/models/Ball.java
+++ b/src/main/java/core/basesyntax/models/Ball.java
@@ -1,0 +1,18 @@
+package core.basesyntax.models;
+
+import core.basesyntax.models.enums.Color;
+
+public class Ball {
+    private Color color;
+    private byte number;
+
+    public Ball(Color color, byte number) {
+        this.number = number;
+        this.color = color;
+    }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/models/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+    private byte number;
+
+    public Ball(Color color, byte number) {
+        this.number = number;
+        this.color = color;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("Color - %s; Number - %d;", this.color.name(), number);
```
KseniiaMakarova, line 16: do we need `this`?

KostyaNikiforov: Why not, it looks pretty

KseniiaMakarova (changes requested):
travis failed
also, let's remove `this` keyword where we don't need it - it reduces readability

KseniiaMakarova (approved):

PR #33, opened 2021-10-20
Lottery was implemented

MTWIB:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
...
-        // create three balls using class Lottery and print information about them in console
+        Ball firstBall = new Lottery().getRandomBall();
+        Ball secondBall = new Lottery().getRandomBall();
+        Ball thirdBall = new Lottery().getRandomBall();
+        System.out.println("First ball. Color - " + firstBall.getColor() + ", number - "
+                + firstBall.getNumber() + ".");
+        System.out.println("Second ball. Color - " + secondBall.getColor() + ", number - "
+                + secondBall.getNumber() + ".");
+        System.out.println("Third ball. Color - " + thirdBall.getColor() + ", number - "
+                + thirdBall.getNumber() + ".");
```
KseniiaMakarova, line 13: let's override `toString()` method in `Ball` class where we will return information about balls (remove `first`, `second` words)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball firstBall = new Lottery().getRandomBall();
+        Ball secondBall = new Lottery().getRandomBall();
+        Ball thirdBall = new Lottery().getRandomBall();
```
KseniiaMakarova, line 7: what if you were asked to create 50 balls?
let's create Lottery object only once
and then use a loop

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
I would recommend to use singular for enum names

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
+        return ball;
```
KseniiaMakarova, line 8:
```suggestion
        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        int i = 0;
+        Ball ball = new Ball();
+        while (i < 3) {
```
KseniiaMakarova, line 7: `3` should be a constant field because it is a magic number

also, use `for` loop instead

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
+            ball = new Lottery().getRandomBall();
```
KseniiaMakarova, line 8: you are creating a `Lottery` object 3 times

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball = new Ball();
```
Sofasmile, line 5: why do we create the ball before the loop? can we avoid it?

PR #37, opened 2021-10-20
My solution

Eugene-exe:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
+    H_OTPINK,
+    E_AGLECOLOR,
+    L_INEN,
+    P_URPLE,
+    M_AGENTA,
```
Sofasmile, line 8: why do we use underline after the first letter?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
+        return ball;
+    }
+
```
Sofasmile, line 15: redundant empty line, we should avoid empty lines after last method in the class

Sofasmile (changes requested):
let's create main() and do this part of the task
image

PR #38, opened 2021-10-20
The lottery for ball was created

kateryna-deshchinska:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
```
Sofasmile, line 5: let's create a constant NUMBER_OF_BALLS for the 3 number (now 3 is a magic number)
read more about constants here - https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
...
+        for (Ball ball : balls) {
+            ball = Lottery.getRandomBall();
```
Sofasmile, line 7: does this line will set a new ball to the array? try to test it

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
Sofasmile, line 6:
```suggestion
    public String getRandomColor() {
```
we should avoid using `static` with methods

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
Sofasmile, line 6:
```suggestion
    public Ball getRandomBall() {
```
the same here

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    static final int NUMBER_OF_BALLS = 3;
```
Sofasmile, line 4:
```suggestion
    private static final int NUMBER_OF_BALLS = 3;

```
better move this constant to the Application class

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        Ball[] balls = new Ball[Ball.NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
Sofasmile, line 8: let's create `new Lottery ` object before for loop

PR #109, opened 2022-01-14
added Ball, Color, Ball and Lottery class and solved hw

AprioriKS:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+
+    public Ball getRandomBall() {
+        Random random = new Random();
+        return new Ball(random.nextInt(100) + 1, getRandomColor());
```
AlexRozv, line 9: why do you need + 1 here?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+
```
AlexRozv, line 6: let's remove redundant empty line

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
```
AlexRozv, line 4: why is it number of users if we should create balls?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[NUMBER_OF_USERS];
+        for (int i = 0; i < lottery.length; i++) {
+            lottery[i] = new Lottery();
+            System.out.println(lottery[i].getRandomBall());
+        }
```
AlexRozv, line 11: why did you create an array of lotteries? you should create only one lottery instance outside the loop, then instead of array of lotteries you should create an array of balls, inside the loop you should fill your array with random balls and then print all the info about them

AlexRozv (changes requested):

src/main/java/core/basesyntax/Lottery.java
AprioriKS: если 100 оставить то вроде как будет число от 0 до 99, а нам же нужно от 1 до 100

AlexRozv (approved):
well done 👍

PR #115, opened 2022-01-16
added solution to HW Lottery

OlehMytiai:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
AlexRozv, line 5: let's remove this comment

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
...
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
+        Ball secondBall = lottery.getRandomBall();
+        Ball thirdBall = lottery.getRandomBall();
+
```
AlexRozv, line 10: let's remove redundant empty lines

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random RANDOM = new Random();
```
AlexRozv, line 6: it is unnecessary to make it a const

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(100));
```
AlexRozv, line 8: why did you create an instance of ColorSupplier, but use Random directly in the return statement? Let's keep our code consistent and use either one of those approaches. (Or create instances for both ColorSupplier and Random, or use both of them directly in return)

AlexRozv (changes requested):

src/main/java/core/basesyntax/Lottery.java
mytiaiov: Copy that

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
+        Ball secondBall = lottery.getRandomBall();
+        Ball thirdBall = lottery.getRandomBall();
+        System.out.println("1st " + firstBall.toString());
+        System.out.println("2nd " + secondBall.toString());
+        System.out.println("3rd " + thirdBall.toString());
```
AlexRozv, line 11: let's use for loop here to create and print balls

AlexRozv (changes requested):

src/main/java/core/basesyntax/Application.java
mytiaiov: Is it a good practice to create const variable for count of balls?
Example:
`private static final int NUMBER_OF_BALLS = 3;`

AlexRozv (approved):
good job 👍

PR #118, opened 2022-01-17
Returned ball and number

pikador10:
Implemented lottery successfully

### DIFF
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(101));
+        return ball.toString();
```
AlexRozv, line 10: this method should return ball, not string

AlexRozv (changes requested):

src/main/java/core/basesyntax/Lottery.java
pikador10: Corrected return type of method getRandomBall

AlexRozv (approved):
well done 👍

PR #120, opened 2022-01-17
hw solution

sltnsrh:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery firstBall = new Lottery();
+        Lottery secondBall = new Lottery();
+        Lottery thirdBall = new Lottery();
```
AlexRozv, line 7: let's create only one instance of lottery and use it to create random balls

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
+
+        System.out.println(firstBall.getRandomBall());
+        System.out.println(secondBall.getRandomBall());
+        System.out.println(thirdBall.getRandomBall());
```
AlexRozv, line 11: let's use for loop to create random balls

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values() [index]);
```
AlexRozv, line 8:
```suggestion
        return String.valueOf(Colors.values()[index]);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100)).toString();
+    }
```
AlexRozv, line 8: this method should return Ball, not a String

AlexRozv (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
sltnsrh: Can`t understand why: "return null;"  
We need return color value? (Colors.values()[index])

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 1; i <= 3; i++) {
+            Lottery ball = new Lottery();
```
AlexRozv, line 6: Let's create Lottery instance outside the loop and call it lottery instead of ball

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #122, opened 2022-01-18
Two classes with methods were added, enum was created

Liubov-Chernohor:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = new Lottery().getRandomBall();
```
AlexRozv, line 6: let's create Lottery instance outside the loop and use it to create random balls

AlexRozv (changes requested):

PR #124, opened 2022-01-18
added solution

yurii-tomashevskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery1 = new Lottery();
+        System.out.println("Ball number 1:" + lottery1.getRandomBall().toString());
+        Lottery lottery2 = new Lottery();
+        System.out.println("Ball number 2:" + lottery2.getRandomBall().toString());
+        Lottery lottery3 = new Lottery();
+        System.out.println("Ball number 3:" + lottery3.getRandomBall().toString());
```
AlexRozv, line 10: 1. let's create only one lottery instance
2. let's use for loop to create and print random balls
3. no need to call toString() method on random balls, because System.out.println automatically calls this method

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
+        Color[] values = Color.values();
```
AlexRozv, line 8: why do you have to create an array of colors if you can access enum directly?

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #125, opened 2022-01-18
1-st implementation of jv-lottery task

OleksiyTitov:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+        ball1.toString();
+        ball2.toString();
+        ball3.toString();
```
AlexRozv, line 11: 1. let's use for loop to create and print random balls
2. calling toString() method on balls won't print info about them, let's use System.out.println

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(100));
```
AlexRozv, line 8: why did you create an instance of ColorSupplier, but use Random directly in the return statement? Let's keep our code consistent and use either one of those approaches. (Or create instances for both ColorSupplier and Random, or use both of them directly in return)

AlexRozv (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,11 +3,9 @@
...
         Lottery lottery = new Lottery();
-        Ball ball1 = lottery.getRandomBall();
-        Ball ball2 = lottery.getRandomBall();
-        Ball ball3 = lottery.getRandomBall();
-        ball1.toString();
-        ball2.toString();
-        ball3.toString();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.printf(ball.toString());
```
AlexRozv, line 8: it's unnecessary to call toString() method on ball because System.out.println does it automatically

AlexRozv (changes requested):

AlexRozv (approved):
well done 👍

PR #129, opened 2022-01-19
Added method in Lottery class which is creating Balls with random par…

karma-o:
…ameters

+ 2 classes(Lottery, Ball) + 1 Enum(Color);
+ declaration of 3 random Balls added to main method along with printing them

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = Lottery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball(Color c, int number) {
+        this.color = c;
+        this.number = number;
+    }
+
+    public String toString() {
+        return this.color.toString() + " " + this.number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
+        return Color.values()[new Random().nextInt(Color.values().length)];
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        return new Ball(
+                ColorSupplier.getRandomColor(),
+                new Random().nextInt(100)
+        );
+    }
+}
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    BLACK,
+    WHITE,
+    GREEN,
+    YELLOW,
+    BLUE,
+    ORANGE,
+    PINK
+}


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
```
AlexRozv, line 6: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-use-static-methods-in-your-solution

AlexRozv: this method shouldn't be static and it should return String

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
```
AlexRozv, line 4: you should save color as a String

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
AlexRozv, line 6: mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-use-static-methods-in-your-solution

AlexRozv (changes requested):

karma-o:
Also I know about the static methods in Lottery and ColorSupplier and that this is a bad practice in general, but honestly, I think it should be there because we are not manipulating any data of the objects with those methods and using them only to get randoms, so there is no need to create objects of those classes that I mentioned above.

src/main/java/core/basesyntax/Ball.java
karma-o: then why would I need enum?

src/main/java/core/basesyntax/Lottery.java
karma-o: I know about the static methods in Lottery and ColorSupplier and that this is a bad practice in general, but honestly, I think it should be there because we are not manipulating any data of the objects with those methods and using them only to get randoms, so there is no need to create objects of those classes that I mentioned above.

src/main/java/core/basesyntax/Ball.java
karma-o: I am using only Enum in every method and there is nothing about that in readme file, only create Ball class with fields color and number.
[README.md](https://github.com/mate-academy/jv-lottery/files/7896420/README.md)

src/main/java/core/basesyntax/Lottery.java
karma-o: I can easily change this to non-static, but still wanted to clarify

src/main/java/core/basesyntax/Ball.java
AlexRozv: Screenshot 2022-01-19 at 13 08 27

getRandomColor method must return String as written in the task description. Now, if we receive random color in String format it's easier to save it as a String, than converting to some other type again. And enum is a good way to store some constant values.

src/main/java/core/basesyntax/Lottery.java
AlexRozv: Let's stick to the suggested guidelines and if you want to go deeper on why is it a bad practice to use static methods you can ask this question in tomorrow's lesson.

AlexRozv (changes requested):

src/main/java/core/basesyntax/Lottery.java
AlexRozv: I believe Bohdan will be able to answer it better than I can.

karma-o: > Let's stick to the suggested guidelines and if you want to go deeper on why is it a bad practice to use static methods you can ask this question in tomorrow's lesson.

ok, no problems.

src/main/java/core/basesyntax/Ball.java
karma-o: > Screenshot 2022-01-19 at 13 08 27
> 
> getRandomColor method must return String as written in the task description. Now, if we receive random color in String format it's easier to save it as a String, than converting to some other type again. And enum is a good way to store some constant values.

oops, didn`t notice that, only looked at this: 

> create Ball class with the following fields: color and number and override toString() method

and there was nothing about type of that variables, so I was choosing myself.
last commit fixes everything.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
AlexRozv, line 7: let's create a Lottery object outside the loop, because now it will be created 3 times and we don't need that

AlexRozv (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery l = new Lottery();
```
AlexRozv, line 5: let's use more informative naming: Lottery l -> Lottery lottery

AlexRozv (approved):
well done 👍

PR #132, opened 2022-01-19
implement lottery app

anyutamysko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] lotteryBalls = new Ball[3];
+        for (int i = 0; i < lotteryBalls.length; i++) {
+            lotteryBalls[i] = new Lottery().getRandomBall();
```
AlexRozv, line 7: let's create Lottery object outside the loop, because now it will be created three times

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Locale;
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[index]).toLowerCase(Locale.ROOT);
```
AlexRozv, line 9:
```suggestion
        return String.valueOf(Colors.values()[index]).toLowerCase();
```
I dont think we need it here, don't forget to remove import for Locale

AlexRozv (changes requested):

AlexRozv (approved):
good job 👍

PR #135, opened 2022-01-19
implement get random ball and printed it

IgorZima90:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,7 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall().toString());
```
AlexRozv, line 6: you need to do it three times, let's use for loop for it

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
```
AlexRozv, line 6: Screenshot 2022-01-19 at 13 08 27
from task description this method should return String and not color, let's fix that.

AlexRozv (changes requested):

AlexRozv (commented):
well done 👍

AlexRozv (approved):

PR #149, opened 2022-02-10
added jv-lottery solution

dmnichz:


MaliukDaria (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
MaliukDaria, line 7: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes): Don’t create several instances of the same class if you can use only one instance for your purposes

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].toString();
```
MaliukDaria, line 7: lets make Random as class level variable, not to create new instanse every time, when we call getRandomColor() method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
MaliukDaria, line 7: same here, lets mare ColorSupplier and Random class level variables

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #151, opened 2022-02-15
implemented task

tanya-logatskaya:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
```
MaliukDaria, line 6: checklist: Don’t create several instances of the same class if you can use only one instance for your purposes[](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)
```suggestion
            System.out.println(new Lottery().getRandomBall());
```

we don't need to explicitly call the toString() method here, this happens under the hood

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
MaliukDaria, line 7: lets make Random class level variable (not local variable) so we won't create new instance of Random class every time we call getRandomColor() method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
MaliukDaria, line 9: same here, lets make ColorSupplier and Random class level variables

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random randomBall = new Random();
```
```suggestion
    private Random random = new Random();
```

not informative variable name

src/main/java/core/basesyntax/Lottery.java
MaliukDaria: what about this comment?

MaliukDaria (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
+        return ball;
+    }
```
MaliukDaria, line 13:
```suggestion
public class Lottery {
	private ColorSupplier colorSupplier = new ColorSupplier();
        private Random random = new Random();
        
    public Ball getRandomBall() {
        Ball ball = new Ball();
        ball.setColor(colorSupplier.getRandomColor());
        ball.setNumber(random.nextInt(100));
        return ball;
    }
```

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #159, opened 2022-02-20
Develop

katy-mozharovska:


okuzan (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
+
+        return Colors.values()[index].toString();
```
okuzan, line 9:
```suggestion
    private Random random = new Random();
    
    public String getRandomColor() {
        int index = random.nextInt(Colors.values().length);
        return Colors.values()[index].name();
```

okuzan: as mentioned in checklist, let's avoid static methods, and calling .name() on enum is preferable in comparison with .toString(). Random - same as in Lottery class
image
image

okuzan (changes requested):

okuzan (approved):
LGTM

PR #163, opened 2022-03-09
Lottery implementation

irynahrynda:
- implemented logic of lottery that returns balls with a random color and random number
- added class Ball
- added enum Color with different colors
- added class Lottery

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(),
+                new Random().nextInt(100));
```
MaliukDaria, line 8: Same here lets make Random and ColorSupplier as class level variables, not to create new instances  every time we call getRandomBall() method

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #165, opened 2022-03-13
Implement lottery logic

dSkarzh:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Ball {
+    private final Color color;
+    private final int number;
+
+    protected Ball(Color color, int number) {
```
KseniiaMakarova, line 7: why protected? the usage of protected is very limited in practice, let's better be not afraid to use public

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    protected Random random = new Random();
+
+    protected Color getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index];
     }
```
KseniiaMakarova, line 11: same here about protected

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
+    protected ColorSupplier colorSupplier = new ColorSupplier();
+    protected Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(BOUND));
+    }
```
KseniiaMakarova, line 12: great job on constants and working with fields in service classes

if you want to go 1 step further, you can pass `new ColorSupplier()` and `new Random()` objects into the Lottery's constructor from `main()`

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Ball.java
dSkarzh: are you talking about this place specificly or about access modifiers in whole project?

src/main/java/core/basesyntax/Lottery.java
dSkarzh: Thanks! Just commited proposed changes

src/main/java/core/basesyntax/Ball.java
KseniiaMakarova: in whole project

KseniiaMakarova (approved):

PR #173, opened 2022-03-15
Hw lottery

Anastasia-702:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
KseniiaMakarova, line 6: don't make methods static
you can call them through the object of a class

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public static String getRandomBall() {
+        Random random = new Random();
+        int value = random.nextInt(100);
+        return getRandomColor() + " " + value;
```
KseniiaMakarova, line 9: you should return a Ball

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int number = random.nextInt(100);
```
KseniiaMakarova, line 8: 100 would look better as a constant field

KseniiaMakarova (approved):

PR #183, opened 2022-03-21
Jv lottery solution

losnazar:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
```
MaliukDaria, line 8: same lets make Random and ColorSupplier as class level variables, not to create new instances every time we call getRandomBall() method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        int number = random.nextInt(100);
+        Ball ball = new Ball(colorSupplier.getRandomColor(),number);
+        return ball.toString();
+    }
+
```
MaliukDaria, line 13: redundant empty line

MaliukDaria (changes requested):

MaliukDaria (approved):
👍🏻

PR #199, opened 2022-03-24
added methods getRandomBall, getRandomColors

xArthurZix:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] allBalls = new Ball[3];
```
Sofasmile, line 5: do we use this array? can we remove it?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
```
Sofasmile, line 5: we should have setters or constructor with parameters for these fields

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index];
+    }
+
+    public int getNumber() {
+        Random random = new Random();
+        int value = random.nextInt(100);
+        return value;
     }
```
Sofasmile, line 15: this method shouldn't belong to the ColorSupplier class because it doesn't relate to colors.

let's generate numbers directly in the getRandomBall method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public class Lottery extends ColorSupplier {
+    public String getRandomBall() {
```
Sofasmile, line 4:
```suggestion
    public Ball getRandomBall() {
```
this method should return Ball object

Sofasmile (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+        int number = random.nextInt(100);
```
Sofasmile, line 9: 100 should be made a constant field (private static final field with name in UPPER_CASE)

Sofasmile (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_NUMBER = 100;
```
Sofasmile, line 6:
```suggestion
    private static final int MAX_NUMBER = 100;
```

Sofasmile (approved):

PR #200, opened 2022-03-24
Implement `getRandomBall` application

lexzane:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
+    private final int number;
```
Sofasmile, line 5: не обов'язково робити ці поля фінальними. final варто використовувати лише якщо для цього є потреба

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Sofasmile, line 7: 100 should be made a constant field (private static final field with name in UPPER_CASE)

Sofasmile (changes requested):

src/main/java/core/basesyntax/Lottery.java
lexzane: Ok, but in this case `nextInt()` method don't return random number of Ball as it's needed

Sofasmile (approved):

PR #212, opened 2022-03-25
Hw lottery solution

vitalii-smahlenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
Nick97-git, line 7: let's make a global field for Random instance and call it, so we don't need to create a new instance every time

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,7 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
```
Nick97-git, line 6: Screenshot 2022-03-28 at 10 20 46

Nick97-git (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Nick97-git, line 6: it's better to create a constant of 3

Nick97-git (approved):

PR #220, opened 2022-04-03
1st try. Working MVP. No code refactoring so far.

drapail:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        System.out.println(ball1);
+        Ball ball2 = Lottery.getRandomBall();
+        System.out.println(ball2);
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball3);
```
MaliukDaria, line 11: lets make this in a loop

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Colors getRandomColor() {
```
MaliukDaria, line 6: lets make this method not static

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        int index = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
MaliukDaria, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
...
+    Red,
+    Orange,
+    Yellow,
+    Green,
```
MaliukDaria, line 7:
```suggestion
    RED,
    ORANGE,
    YELLOW,
    GREEN,
```

we need to use CAPS for Enums elements

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
MaliukDaria, line 6: lets make this method not static

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        int randomNumber = new Random().nextInt(101);
+        Ball ball = new Ball(ColorSupplier.getRandomColor(), randomNumber);
```
MaliukDaria, line 8: same lets make Random and ColorSupplier as class level variables, not to create new instances every time we call getRandomBall() method. And lets move 101 to constant https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

MaliukDaria (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
```
MaliukDaria, line 7: [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes): Don’t create several instances of the same class if you can use only one instance for your purposes

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int UPPER_LIMIT = 101;
```
MaliukDaria, line 6: what about an access modifier?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+    private static Random random = new Random();
+
+    public Ball getRandomBall() {
+        int randomNumber = new Random().nextInt(UPPER_LIMIT);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
MaliukDaria, line 11: same lets make ColorSupplier as class level variable, not to create new instances every time we call getRandomBall() method

MaliukDaria (changes requested):

src/main/java/core/basesyntax/Lottery.java
drapail: I didn't get this one, should this be private instead of static?

MaliukDaria: static it os not access modifier. It should be `private static final int`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static Random random = new Random();
```
```suggestion
    private Random random = new Random();
```

Why static?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private final int upperLimit = 101;
```
MaliukDaria, line 8:
```suggestion
    private Random random = new Random();
    private ColorSupplier colorSupplier = new ColorSupplier();
    private static final int UPPER_LIMIT = 101;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(colorSupplier.getRandomColor(), random.nextInt(upperLimit));
+        return ball;
```
MaliukDaria, line 12:
```suggestion
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(upperLimit));
```

MaliukDaria (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
drapail: I thought it was needed so that I am able to call method without instantiating an object. Now I see that I never actually used this feature and used the instance instead

MaliukDaria (approved):
👍🏻

PR #235, opened 2022-04-19
Task is done.

ChumakAndrey:


KostyaNikiforov (changes requested):
Let's replace all static methods with ordinary ones according to the checklist
Screenshot 2022-05-16 at 10 06 51

PR #236, opened 2022-04-21
created Ball, Lottery, classes; created enum Color, implemented getRa…

viktoriia-haraniuk:
…ndomColor(), getRandomBall() methods; tested by creating 3 balls and printing theirs information

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < 3; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private String colour;
+    private int number;
+
+    public Ball(String colour, int number) {
+        this.colour = colour;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball's color: " + colour + "; number: " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
```
KostyaNikiforov, line 6: Can we get rid of this array?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
KostyaNikiforov, line 10: You can use method name() to get the name of the enum element

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BOUND = 100;
+    private ColorSupplier supplier = new ColorSupplier();
```
KostyaNikiforov, line 7: This field can be final too

KostyaNikiforov (changes requested):
This solution is okay, but to make it perfect, you can pass dependencies to the constructor in the class ColorSupplier and Lottery.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -2,12 +2,14 @@
 
 import java.util.Random;
 
-public class Lottery {
+public class Lottery extends ColorSupplier {
```
KostyaNikiforov, line 5: In this case, It is better to use composition instead of inheritance

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public class Ball {
+    private String colour;
```
KostyaNikiforov, line 4:
```suggestion
    private String color;
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball's color: " + getColour() + "; number: " + getNumber();
+    }
```
KostyaNikiforov, line 31: Inside the class, you can get access to its fields directly

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
Good job)

PR #247, opened 2022-04-30
Hw jv lottery

zim10093:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int number;
+
+    public Ball(Colors color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{" + "COLOR=" + color + ", number=" + number + '}';
```
KostyaNikiforov, line 14: It is a good practice to use the next format to implement toString method:
```suggestion
        return "Ball{" 
            + "color=" + color 
            + ", number=" + number 
            + '}';
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
KostyaNikiforov, line 5: Each time this method is called, a new ColorSupplier instance will be created. Let's think about how we can optimize this logic and create an instance only once.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -2,6 +2,6 @@
 
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[(int) (Math.random() * Colors.values().length)].toString();
```
KostyaNikiforov, line 5: It would be cool to move some logic out of this line to increase the readability. 
Remember about private methods ;)

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -1,8 +1,9 @@
 package core.basesyntax;
 
 public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
```
KostyaNikiforov, line 4: This solution is okay, but to make it perfect, you can pass dependencies to the constructor.

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #249, opened 2022-05-02
Jv lottery

Ivan-Yuriv:
The lottery with balls was created.

### DIFF
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int value = new Random().nextInt(100);
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int value = new Random().nextInt(100);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
KostyaNikiforov, line 8: Each time this method is called, new ColorSupplier and Random instanced are created. Let's think about how we can optimize this logic and create each instance only once.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int value = new Random().nextInt(100);
```
KostyaNikiforov, line 7: Why are you using the number 100 here?
To answer this question to someone who will read your code, you can make this value constant.

KostyaNikiforov (changes requested):

PR #251, opened 2022-05-04
Complete Lottery task

Oleg-pyt:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = {new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()};
```
Sofasmile, line 7: let's create a separate variable for `Lottery` instance and use it:
```
```
and better call method `getRandomBall()` in the loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Sofasmile, line 7: let's create a constant for 100 - https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

Sofasmile (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
...
+        for (Ball ball: balls) {
+            System.out.println(ball.toString());
```
Sofasmile, line 9: it's optional to call toString() method
it will be called by default

src/main/java/core/basesyntax/Lottery.java
Sofasmile, line 7: be attentive, I've written about constant for 100 number

Sofasmile (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,7 +3,9 @@
 import java.util.Random;
 
 public class Lottery {
+    public static final int MAX_NUMBER = 100;
```
Sofasmile, line 6: if we use this constant only in this class we can make it private

Sofasmile (approved):

PR #257, opened 2022-05-10
Completed task from file README

Pavlo99-gif:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+public class Ball {
+    private Colors color;
```
KostyaNikiforov, line 4:
```suggestion
    private Color color;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
```
KostyaNikiforov, line 6: Each time this method is called, new ColorSupplier and Random instanced are created. Let's think about how we can optimize this logic and create each instance only once.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
KostyaNikiforov, line 9: Let's make this value constant, as the checklist require)

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):
This solution is okay, but to make it perfect, you can pass dependencies to the constructor.
Let's strive for perfection!)

KostyaNikiforov (changes requested):
Check the last message in the group chat)

KostyaNikiforov (approved):
Good job)

PR #267, opened 2022-05-16
hw lottery

TymurTsvietkov:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[index]);
```
KostyaNikiforov, line 8: You can use `name()` method to get the name of the enum element

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
+        int number = random.nextInt(100);
```
KostyaNikiforov, line 9: et's make this value constant, as the checklist require)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
```
KostyaNikiforov, line 8: Check my messages in the group chat)
We would avoid creating a lot of redundant objects

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,10 +3,9 @@
...
 public class Lottery {
-    public static Ball getRandomBall() {
-        ColorSupplier color = new ColorSupplier();
-        Random random = new Random();
-        int number = random.nextInt(100);
-        return new Ball(color.getRandomColor(), number);
+    public static final int NUMBER = 100;
+
+    public static Ball getRandomBall(Random random, ColorSupplier color) {
+        return new Ball(color.getRandomColor(), random.nextInt(NUMBER));
```
KostyaNikiforov, line 9: Don't pass the class dependencies to the method params, it is better to use the constructor

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -5,6 +5,6 @@
 public class ColorSupplier {
     public String getRandomColor() {
         int index = new Random().nextInt(Colors.values().length);
```
KostyaNikiforov, line 7: The same with the Random instance

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER = 100;
+    private final Random random;
+    private final ColorSupplier color;
```
KostyaNikiforov, line 8:
```suggestion
    private final ColorSupplier colorSupplier;
```

KostyaNikiforov (approved):

PR #268, opened 2022-05-17
Ball lottery implemented

antonEEmus:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
KostyaNikiforov, line 5: Remove redundant comment)

KostyaNikiforov (changes requested):
Check the checklist one more time, please) There is a point about static methods.
Also, check my messages from the group chat and make a refactor of your solution according to this principle to make it perfect)

KostyaNikiforov (approved):
:fire:

PR #272, opened 2022-05-18
Lotery

ill4ik:


ill4ik:
first commit (my solution is not correct) ,please check second commit (my correct solution)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
+        Lottery lot = new Lottery();
+        Ball ball1 = lot.getRandomBall();
+        Ball ball2 = lot.getRandomBall();
+        Ball ball3 = lot.getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
KostyaNikiforov, line 11: Can we use a loop here?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random r = new Random();
+        int value = r.nextInt(100);
```
KostyaNikiforov, line 8: Let's make this value constant, as the checklist requires)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
```
KostyaNikiforov, line 6: Each time this method is called, new instances of ColorSupplier and Random are created. Let's think about how we can optimize this logic and create each instance only once.

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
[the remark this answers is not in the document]
KostyaNikiforov: What about this random instance?)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lot = new Lottery();
+        Ball ball1 = lot.getRandomBall();
+        Ball ball2 = lot.getRandomBall();
+        Ball ball3 = lot.getRandomBall();
```
KostyaNikiforov, line 8: I guess we can combine this into a loop too

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Lottery(Random random, ColorSupplier colorSupplier) {
+        this.colorSupplier = colorSupplier;
+        this.random = random;
+    }
+
+    public Ball getRandomBall() {
+        final int value = random.nextInt(100);
```
KostyaNikiforov, line 15: This value should be a constant

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):
🔥

PR #273, opened 2022-05-18
Lottery homework

NikitaSemisoshenko:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random randomBall = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(randomBall.nextInt(100));
```
KostyaNikiforov, line 11: Each time this method is called, new instances of ColorSupplier and Random are created. Let's think about how we can optimize this logic and create each instance only once.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        Random randomColor = new Random();
```
KostyaNikiforov, line 8: The same here
Let's pass the objects to the constructor as I suggested in the message in the group chat

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
KostyaNikiforov, line 3:
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+    private final Random random;
+
+    public Lottery(ColorSupplier colorSupplier, Random random) {
+        this.colorSupplier = colorSupplier;
+        this.random = random;
+    }
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(random.nextInt(100));
```
KostyaNikiforov, line 16: The magic number 100 should be a [constant](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Colors.java
KostyaNikiforov: What about the name of enum? It must be singular

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    private static final int RANDOM_INTEGER = 100;
```
KostyaNikiforov, line 6:
```suggestion
    private static final int MAX_RANDOM_INTEGER = 100;

```

KostyaNikiforov (approved):

PR #274, opened 2022-05-18
Tasks compleded, build success

oleksii-kordysh:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] randomBall = new Ball[3];
```
KostyaNikiforov, line 6: Can we get rid of this array?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball randomBall = new Ball();
+        ColorSupplier randomColor = new ColorSupplier();
```
KostyaNikiforov, line 8: Each time this method is called, new instances of ColorSupplier and Random are created. Let's think about how we can optimize this logic and create each instance only once.

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Application.java
oleksii-kordysh: My misunderstanding) Its not neccessery

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    static final int COLORS_AMOUNT = Colors.values().length;
+    private Random random = new Random();
```
KostyaNikiforov, line 7: Just like this
```suggestion
    private Random random;
    
    public ColorSupplier(Random random) {
        this.random = random;
    }
```

KostyaNikiforov (changes requested):
Your implementation is almost perfect but we can make it even better if we pass random and colorSupplier to the constructor

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier(random);
+
```
KostyaNikiforov, line 8:
```suggestion
     private final Random random;
     private final ColorSupplier colorSupplier;

      public Lottery(Random random, ColorSupplier colorSupplier) {
          this.random = random;
          this.colorSupplier = colorSupplier;
       }
```

KostyaNikiforov (changes requested):

PR #278, opened 2022-05-18
Created Lottery

DimaPodakin:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+    private Random random;
+
+    public Lottery(ColorSupplier colorSupplier, Random random) {
+        this.random = random;
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
+        return ball;
```
KostyaNikiforov, line 16: Let's make the value 100 constant, as the checklist requires)
And also we can combine it into one line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
KostyaNikiforov, line 7: We can pass Random object to the constructor here too

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[index].toString();
```
KostyaNikiforov, line 8: (obtional) You can use name() method to get the name of the enum element

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public enum Colors {
```
KostyaNikiforov, line 3:
```suggestion
public enum Color {
```

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,6 +3,7 @@
 import java.util.Random;
 
 public class Lottery {
+    private static int Value = 100;
```
KostyaNikiforov, line 6: [constant-names](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/core/basesyntax/Colors.java
KostyaNikiforov: Enum name must be singular

src/main/java/core/basesyntax/ColorSupplier.java
KostyaNikiforov: Just as you do it in the Lottery class

KostyaNikiforov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int VALUE = 100;
```
KostyaNikiforov, line 6:
```suggestion
    private static final int MAX_BALL_NUMBER = 100;
```

KostyaNikiforov (approved):

PR #291, opened 2022-05-26
add solution

uglovyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
```
Nick97-git, line 8: use for loop and create variable for Lottery before for loop, so you won't need to create a new instance for every ball

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Random;
+
+public class Ball {
+    private final String color;
+    private final int number;
+
+    public Ball() {
+        this.color = new ColorSupplier().getRandomColor();
+        this.number = new Random().nextInt(100);
```
Nick97-git, line 11: logic of getting of random color and number, should be move to Lottery's method, in constructor you should pass already formed parameters to initialize your object

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[(new Random().nextInt(Colors.values().length - 1))].name();
```
Nick97-git, line 7: create a global Random field and use it here

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[(random.nextInt(Color.values().length - 1))].name();
```
MaliukDaria, line 9:
```suggestion
        return Color.values()[(random.nextInt(Color.values().length))].name();
```

why `length - 1`?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
MaliukDaria, line 7: same, lets make Random and ColorSupplier as class level variables, not to create new instances every time we call getRandomBall() method. Also lets make number 100 as constant field https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

MaliukDaria (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public class Ball {
+    static final int MAX_NUMBER = 100;
```
MaliukDaria, line 4: I think It's better to move this constant in Lottery class(as we can have many lotteries in the future with different max numbers). And don't forget about access modifiers(if you don't use this constant outside class, where it defined, make it private)

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #293, opened 2022-05-31
Add classes Ball, Color, Lottery

voitalik:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
MaliukDaria, line 8:
```suggestion
            System.out.println(lottery.getRandomBall());
```

`toString()` method will be called "under the hood"

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int number = new Random().nextInt(100);
+        String randomColor = new ColorSupplier().getRandomColor();
```
MaliukDaria, line 8: same, let's make Random and ColorSupplier as class level variables, not to create new instances every time we call getRandomBall() method. And lets make number 100 as constant field https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
```suggestion
    public Color getRandomColor() {
```

if you use Color in class Ball lets return Color from `getRandomColor()` method

MaliukDaria (changes requested):

MaliukDaria (approved):
GJ🔥

voitalik:
What should I do after approval? Close request or leave it as is?

PR #321, opened 2022-06-21
fix maven error

Oldudka:


Nick97-git (changes requested):
you should create PR not from master, but a new branch
Screenshot 2022-06-21 at 13 38 35

PR #326, opened 2022-06-24
Lottery

VolodymyrBarik:


Nick97-git (changes requested):
I don't see package (screenshot 2 - current package structure) structure like it was, seems like you removed packages core/basesyntax (screenshot 1), return them
Screenshot 2022-06-24 at 17 02 44
Screenshot 2022-06-24 at 17 03 07

PR #335, opened 2022-07-02
Maven throws checkstyle errors, in particular, the File (package core…

oleksandr-gotin:
….basesyntax) does not end with a newline. [NewlineAtEndOfFile]


Nick97-git (changes requested):
fix this first
Screenshot 2022-07-04 at 08 59 06

Nick97-git (changes requested):
you should've added your solution in a separate branch not master
Screenshot 2022-07-04 at 20 15 28

PR #339, opened 2022-07-04
task solution_23:30

oleksandr-gotin:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
```
```suggestion
    private static final int MAX_RANDOM_NUMBER = 100;
```

Nick97-git (changes requested):
it's a bad commit message, you should describe in short form what changes you do in commits, you can't fix this one, but for the future, think about what you write here
Screenshot 2022-07-05 at 08 06 31

PR #340, opened 2022-07-05
Implemented java-lottery

oleh-satoshi:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return color + " ---> " + number;
```
Noct2000, line 14: I think it is not the best suggestion
```suggestion
        return color + " ---> " + number;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int value = random.nextInt(100);
```
Noct2000, line 8: use constant for 100
```suggestion
        int value = random.nextInt(100);
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + getColor() + ", " + "number: " + getNumber();
```
Noct2000, line 30: you can use color instead of getColor() in the toString() method
```suggestion
    public String toString() {
        return "Color: " + getColor() + ", " + "number: " + getNumber();
```

Noct2000 (approved):

PR #354, opened 2022-07-10
task solving

Litvynets:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
MaliukDaria, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,25 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+
+    }
```
MaliukDaria, line 9: do we use this constructor?

MaliukDaria (approved):
GJ!

PR #358, opened 2022-07-13
added mock solution

RomanDubovskyi:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -2,6 +2,7 @@
 
 public class ColorSupplier {
     public String getRandomColor() {
+        // This is my solution
```
RomanDubovskyi, line 5: This is wrong

aliceblunt8 (changes requested):
rejected

aliceblunt8 (approved):

PR #375, opened 2022-07-25
Tree balls was created, their color and number was printed

ghost:
Is it correct to use "System.out.println(lottery.getRandomBall())"? Or is it better to create balls first?

### DIFF
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        int number = random.nextInt(100);
+        String color = colorSupplier.getRandomColor();
+        return new Ball(color, number);
+    }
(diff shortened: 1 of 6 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        int number = random.nextInt(100);
```
Noct2000, line 9: replace 100 with constant 
```suggestion
        int number = random.nextInt(100);

```

Noct2000 (changes requested):

src/main/java/core/basesyntax/Lottery.java
MyroslavVoskoboinyk: I made constant (MAX_NUMBER) in Lottery class.
In witch class better to do this constant?
Beacose, i think that this constant discibe state of Ball(s). Maybe, it was be better made this constant in Ball class?

Noct2000 (approved):
nice work

PR #376, opened 2022-07-25
created classes: Ball and Lottery, enum Colors, implemented methods g…

ArturZhadan:
…etRandomColor() and getRandomBall()

### DIFF
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
+    RED, BLUE, YELLOW, GREEN, BLACK, GRAY, PINK
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Noct2000, line 7: better create ColorSupplier field and define it in Lottery constructor, after that, you can create a new ColorSupplier in the main method and give it to your Lottery instance
PS. replace 100 with constant 
```suggestion
        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));

```

Noct2000 (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {

```

Noct2000 (changes requested):

Noct2000 (approved):

PR #380, opened 2022-07-26
realized task is to return three random balls

SofiiaKazanivska:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
zachar-chorny, line 7:
```suggestion
    private final Random random = new Random();
    
    public static String getRandomColor() {
        int index = random.nextInt(Color.values().length);
```
It is better to make the Random object a class field so that it does not have to be created every time you call a method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int maxRandomNumber = 100;
```
zachar-chorny, line 6: this is not a constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
zachar-chorny, line 9: Do the same here with ColorSuplier and Random as I showed above

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Color.values()[index].toString();
```
zachar-chorny, line 8:
```suggestion
        return Color.values()[index].name();
```

zachar-chorny (changes requested):

PR #384, opened 2022-07-26
created classes: Lottery, ColorSupplier, Colors(enum class) and Ball.

veers-droid:
Ball have fields: String color and int number. It has a constructor and overriden method to string.
Lottery creates a new Ball class with random number and color(colors we take from ColorSupplier which takes from enum class)

### DIFF
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,18 @@
+package core.basesyntax;
+
+
+public enum Colors {
+    RED,
+    GREEN,
+    BLUE,
+    YELLOW,
+    BROWN,
+    GREY,
+    ORANGE,
+    WHITE,
+    BLACK,
+    PURPLE,
+    GOLD,
+    SILVER,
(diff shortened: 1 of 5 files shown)


Noct2000 (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
```
Noct2000, line 6: fix constant naming
```suggestion
    private static final int maxNumber = 100;
```

Noct2000 (changes requested):

Noct2000 (approved):

PR #391, opened 2022-07-27
created classes and methods to return ball with random values

AndrewHudzenko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
```
Noct2000, line 6: move lottery creation out of the loop
```suggestion
            Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+    public String toString() {
+        return "Your random ball:\n"
+                + "humber - "
+                + number
+                + "\n"
+                + "color - "
+                + color
+                + "\n"
+                + "\n"
+                .trim();
```
Noct2000, line 25: use System.lineSeparator() instead of '\n', and i think .trim() is redundant here
```suggestion
    public String toString() {
        return "Your random ball:\n"
                + "humber - "
                + number
                + "\n"
                + "color - "
                + color
                + "\n"
                + "\n"
                .trim();
```

Noct2000 (changes requested):

Noct2000 (approved):

PR #392, opened 2022-07-27
Hw jv lottery

georgiy-nezbritskiy:


src/main/java/mate/Colors.java
```diff
@@ -0,0 +1,10 @@
+package mate;
+
+public enum Colors {
```
Noct2000, line 3:
```suggestion
public enum Color {
```

src/main/java/mate/Colors.java
```diff
@@ -0,0 +1,10 @@
...
+    yellow,
+    blue,
+    red,
+    black,
+    green,
+    white
```
Noct2000, line 9: use uppercase
```suggestion
    yellow,
    blue,
    red,
    black,
    green,
    white
```

src/main/java/mate/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package mate;
+
+import java.util.Random;
+
+public class Lottery {
+
+    static final int MAX_NUMBER = 100;
```
Noct2000, line 7: add access modifier
```suggestion
    static final int MAX_NUMBER = 100;
```

src/main/java/mate/Main.java
```diff
@@ -0,0 +1,9 @@
+package mate;
+
+public class Main {
+    public static void main(String[] args) {
+        for (int i = 0; i < 3; i++) {
+            System.out.println((i + 1) + " " + new Lottery().getRandomBall().toString());
```
Noct2000, line 6: move lottery creation outside of the loop
```suggestion
        for (int i = 0; i < 3; i++) {
            System.out.println((i + 1) + " " + new Lottery().getRandomBall().toString());
```

Noct2000 (changes requested):

src/main/java/mate/Main.java
georgiy-nezbritskiy: but in "Common mistakes" wrote, "Use for loop for creating several objects of the same class". That's why I did so. 

https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-for-loop-for-creating-several-objects-of-the-same-class

georgiy-nezbritskiy: Shoud I change the solution?

src/main/java/mate/Lottery.java
georgiy-nezbritskiy: Why do I have to add an access modifier to the const?

src/main/java/mate/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package mate;
+
+import java.util.Random;
+
+public class Lottery {
+
```
Noct2000, line 6:
```suggestion
```

Noct2000 (approved):

PR #401, opened 2022-07-27
ready solution

JirkVorl:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
+    }
```
Noct2000, line 8: replace 100 with constant 
```suggestion
        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
    }
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/Lottery.java
JirkVorl: > replace 100 with constant

done

Noct2000 (approved):

PR #405, opened 2022-07-27
lottery_random_ball_hw

Jhuzeppe:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random value = new Random();
+        int num = value.nextInt(100);
```
Noct2000, line 9: replace 100 with constant
```suggestion
        int num = value.nextInt(100);
```

Noct2000 (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random value = new Random();
+        final int limit = 100;
+        int num = value.nextInt(limit);
+        return new Ball(color.getRandomColor().toString(), num);
+    }
+}
```
Noct2000, line 13: if you want to create a constant limit = 100 you should define it like this
```suggestion
public class Lottery {
        private static final int LIMIT = 100;

    public Ball getRandomBall() {
        ColorSupplier color = new ColorSupplier();
        Random value = new Random();
        int num = value.nextInt(limit);
        return new Ball(color.getRandomColor().toString(), num);
    }
}
```

Noct2000 (changes requested):

Noct2000 (approved):
don't use master or main branches for commits!

PR #406, opened 2022-07-27
Implement lottery logic

KosatskiyKirill:


Noct2000 (changes requested):
сheck `mvn clean package` and don't change pom.xml

PR #410, opened 2022-07-27
Create lottery for tree random balls

Aptypio13:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
```
Noct2000, line 7: don't create a new Lottery object every time before calling getRandomBall, and better use loop
```suggestion
        System.out.println(new Lottery().getRandomBall());
        System.out.println(new Lottery().getRandomBall());
        System.out.println(new Lottery().getRandomBall());
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
+    private final int indexRandom = new Random().nextInt(Colors.values().length);
```
Noct2000, line 7: move indexRandom to getRandomColor method, and rename indexRandom to randomIndex or index
```suggestion
    private final int indexRandom = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Noct2000, line 8: replace 100 with constant
```suggestion
        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```

Noct2000 (changes requested):

PR #411, opened 2022-07-27
added solution

Maksym-Korkosh:
created simple lottery with Random class


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int max_number = 100;
```
Noct2000, line 6: fix constant naming

Noct2000 (approved):

PR #430, opened 2022-08-12
task solved

andrii-petruk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].name().toLowerCase();
```
```suggestion
        int index = random.nextInt(Color.values().length);
        return Color.values()[index].name().toLowerCase();
```
create field for Random in class

aaananas (changes requested):

PR #437, opened 2022-08-15
Hw 2.1solution

roman680:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Lottery ball = new Lottery();
```
aaananas, line 6:
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < 3; i++) {
```
this way program will not create unnecessary objects
Pay attention to namings, new Lottery() creates an instance of lottery, not an instance if ball

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        final int randomNumber = new Random().nextInt(101);
+        Ball randomBall = new Ball(new ColorSupplier().getRandomColor(), randomNumber);
+        return randomBall;
```
aaananas, line 9:
```suggestion
public class Lottery {
    private static final ColorSupplier colorSupplier = new ColorSupplier();
    private static final Random random = new Random();
    
    public Ball getRandomBall() {
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(101));
```
would be better to extract objects like this to class fields
don't create a variable if you return the value immediately after assigning to that variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
```suggestion
public class ColorSupplier {
    private static final Random random = new Random();
    public String getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
-        return null;
+        Random random = new Random();
```
aaananas, line 7:
```suggestion
```

aaananas (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
aaananas, line 7:
```suggestion
            System.out.println(lottery.getRandomBall());
```
toString() will be invoked automatically

src/main/java/core/basesyntax/ColorSupplier.java
aaananas: up

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int maxValue = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(
+                new ColorSupplier().getRandomColor(), new Random().nextInt(maxValue));
```
aaananas, line 10:
```suggestion
public class Lottery {
    private static final int MAX_VALUE = 100;
    private static final ColorSupplier colorSupplier = new ColorSupplier();
    private static final Random random = new Random();

    public Ball getRandomBall() {
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE));
```
Constant values should be private static final and in uppercase
Good practice to move objects like ColorSupplier, Random to class fields (later you are gonna learn about it more)

aaananas (changes requested):

PR #444, opened 2022-08-18
jv-lottery

Oleksandr-Radovets:


src/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
```
aaananas, line 8: why have you changed .gitignore, and you shouldn't have pushed files other than files from main/java/core/basesyntax

aaananas (changes requested):

aaananas (changes requested):
you shouldn't have removed or changed .gitignore in the first place
Lets create a new branch from master, move your solution there 
And DON'T touch .gitignore file

PR #468, opened 2022-09-08
Finished and Implemented Lottery program

GHSpeedWagon:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    public String getRandomBall() {
```
hronograf, line 6: When the method has name `getRandomBall` and we have class `Ball` it can be confusing if we return `String`, <br>it would be better if return type would be `Ball` (like in your `getRandomColor` method)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+        ColorSupplier colSup = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setColor(colSup.getRandomColor().toString());
+
+        Random index = new Random();
```
hronograf, line 11: Why do we use name `index` here?
index - is something relative to arrays and other data structures and maybe not to balls

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int VALUE = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < VALUE; i++) {
+            Lottery ball = new Lottery();
```
hronograf, line 8: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
...
+            System.out.println(ball.getRandomBall().toString());
```
hronograf, line 9: Can we just write `System.out.println(ball.getRandomBall());` here?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int VALUE = 3;
```
hronograf, line 4: maybe `VALUE` is not the best name in this case, try to find better one)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public ColorsForBall getRandomColor() {
+        int colorForBallRandom = new Random().nextInt(ColorsForBall.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public Ball getRandomBall() {
+
+        Ball ball = new Ball();
+        ball.setColor(colSup.getRandomColor().name());
+        ball.setNumber(random.nextInt(FIGURE_COUNT));
+
+        return ball;
+
+    }
```
hronograf, line 18: [Please don’t add redundant empty lines to your code.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#please-dont-add-redundant-empty-lines-to-your-code)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int FIGURE_COUNT = 100;
```
hronograf, line 6: maybe we can find better name for `FIGURE_COUNT` constant,
because it isn't obvious what the figures we are talking about and why it is their count

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
```
hronograf, line 5: we can add a constructor to our class, so we won't need to set them
our object will be valid after creation

hronograf (changes requested):
Good!
let's improve this solution even more)

hronograf (approved):
Great! 💥

PR #469, opened 2022-09-11
added classes and implemented methods according to the assignment

igorgalenko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int BALLS_NUMBER = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < BALLS_NUMBER; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
hronograf, line 9: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        return Colors.values()[(new Random().nextInt(Colors.values().length))];
```
hronograf, line 7: We can add field with type `Random` to our class, so we won't need to create an instance every time

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(MAX_NUMBER + 1));
```
hronograf, line 9: We can add fields with types `ColorSupplier` and `Random` to our class

hronograf (changes requested):
Good job!

hronograf (approved):
Great!

PR #470, opened 2022-09-12
Jv lottery solution

oksana-rudenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
+    white,
+    black,
+    red,
+    green,
+    blue,
+    yellow,
+    brown
```
MaliukDaria, line 10:
```suggestion
    WHITE,
    BLACK,
    RED,
   ...

```

enum elements should be written using CAPS

MaliukDaria (changes requested):

zachar-chorny (approved):

PR #473, opened 2022-09-16
create Ball class with the following fields: color and number and ove…

maks-untilov:
…rride toString() method

create enum with different colors
implement method getRandomColor() in class ColorSupplier. This method should return a random color, based on the enum values (use Random class for this) Then create a Lottery class with getRandomBall() method, which will return the ball. This method should create a ball with a random color and a random number (the maximum possible number should be 100) and return this ball from the method.


kozhukhovsky (changes requested):

PR #479, opened 2022-09-18
implemented task

vov4ik89:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int i = new Random().nextInt(Colors.values().length);
```
hronograf, line 7: We can add field with type `Random` to our class, so we won't need to create an instance every time when method is invoked

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+public class Lottery {
+    public static final int MAX_VALUE = 100;
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier randomColor = new ColorSupplier();
+        ball.setColor(randomColor.getRandomColor());
+        ball.setNumber(new Random().nextInt(MAX_VALUE + 1));
+        return ball;
+    }
```
hronograf, line 14: We can add fields with type `Random` and `ColorSupplier`  to our class, so we won't need to create an instance every time when method is invoked
Also, we could choose better name for variable with type `ColorSupplier` (like `colorSupplier`)

hronograf (changes requested):
You can do it better)

hronograf (approved):
Cool!😎

PR #483, opened 2022-09-19
lottery logic implementation

shmex-j:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
```
liliia-ponomarenko, line 10: it's bad practice to initialize dependencies in the method. you can initialize it in fields (like in this example ):
image
 but good practice initialization using a constructor:
```
    private final Random random;
    private final ColorSupplier supplier;

    public Lottery(Random random, ColorSupplier supplier) {
        this.random = random;
        this.supplier = supplier;
    }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
liliia-ponomarenko, line 7: the same that in method ```getRandomBall()```

liliia-ponomarenko (changes requested):
Good job 👍  let’s improve your solution!

liliia-ponomarenko (approved):
great job 😎

PR #486, opened 2022-09-20
solution was added

vyshnevsky:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
liliia-ponomarenko, line 6: image

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
liliia-ponomarenko, line 6: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)

liliia-ponomarenko (changes requested):
Let’s fix all common mistakes😉

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class Application {
+    public static final int MAGIC_NUMBER = 3;
```
liliia-ponomarenko, line 6:
```suggestion
    public static final int NUMBER_OF_BALLS = 3;
```

liliia-ponomarenko (approved):
good job but make your constant names more informative 😉

PR #494, opened 2022-09-26
hw-lottery-oop

PavloPolovyi:


liliia-ponomarenko (changes requested):
Let’s  improve your solution 😉

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Color[] colors = Color.values();
+    private final int size = colors.length;
```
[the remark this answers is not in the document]
PavloPolovyi: I think it would be a costly operation to create an array from enum's values every time a method is called, more space complexity and longer execution time. What if enum contains thousands of colors? I thought, it would be better to initialize it once.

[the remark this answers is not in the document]
liliia-ponomarenko: I can agree with you but in your solution you every time create an object  of ``ColorSupplier``` and call the method ```getRandom()``` color only once, so it is the same

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        int numberOfObjectsToCreate = 3;
```
liliia-ponomarenko, line 5: Let’s  make this variable private constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier supplier = new ColorSupplier();
+    private final Random randomizer = new Random();
+    public static final int MAX_NUMBER = 100;
```
liliia-ponomarenko, line 8:
```suggestion
    private static final int MAX_NUMBER = 100;
```
Also, let’s follow the [ordering of class contents](https://mate-academy.github.io/style-guides/java/java.html#s3.4.2-ordering-class-contents)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public Ball getRandomBall() {
+        return new Ball(randomizer.nextInt(MAX_NUMBER + 1), supplier.getRandomColor(randomizer));
```
liliia-ponomarenko, line 11: so hard readability. Let’s create some variables.

liliia-ponomarenko (changes requested):
Let's fix something else. Also, fix problems with JavaCI checks😉

liliia-ponomarenko (approved):
great job! 🔥

PR #495, opened 2022-09-26
Implicated the given task. Lottery return a ball with a random color and number.

YuriiHryniv:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,20 @@
 package core.basesyntax;
 
+import java.util.ArrayList;
+import java.util.List;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        List<Ball> ballList = new ArrayList<>();
+        Lottery lt = new Lottery();
```
hronograf, line 9: long names are good
maybe `lt` is not the best one)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,30 @@
...
+                return Colors.BLACK.name();
+            case 5:
+                return Colors.GREEN.name();
+            case 6:
+                return Colors.SCARLET.name();
+            default:
+                return Colors.BROWN.name();
+        }
     }
 }
```
hronograf, line 30: what if we will add 10 more colors?
1. we should change our `COLOR_COUNT` constant
2. we should add more cases

(`Colors.values()` method may help you to prevent this)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANDOM_NUMBER = 100;
+    private final Random random = new Random();
+    private ColorSupplier cs = new ColorSupplier();
```
hronograf, line 8: programmers like informative names
maybe we can find a better one for `cs`
also, `RANDOM_NUMBER` isn't so _random_, we can find a better name

hronograf (changes requested):
Good job!

YuriiHryniv:
Thanks!

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final int COLOR_COUNT = Colors.values().length;
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int colorNumber = random.nextInt(COLOR_COUNT);
+        return "" + Colors.values()[colorNumber];
```
hronograf, line 11: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)
(`"" + Colors.values()[colorNumber]` here `toString` will be called automatically)

hronograf (changes requested):
Nice!
let's make one tiny fix)

YuriiHryniv:
got it, thx!

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 public class ColorSupplier {
+    private static final int COLOR_COUNT = Colors.values().length;
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int colorNumber = random.nextInt(COLOR_COUNT);
+        Colors targetColor = Colors.values()[colorNumber];
+        return targetColor.name();
     }
```
hronograf, line 13: [Don’t create temporary variables if you can directly return value from the method](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-temporary-variables-if-you-can-directly-return-value-from-the-method)

hronograf (approved):
Great! 💥 
and @YuriiHryniv check the comment)

src/main/java/core/basesyntax/ColorSupplier.java
YuriiHryniv: Okay!

PR #498, opened 2022-09-27
created solution to the lottery task

Bohdan20b:


liliia-ponomarenko (changes requested):
Let’s improve your solution😉

liliia-ponomarenko (approved):
LGTM!🔥

PR #499, opened 2022-09-27
create a simple lottery that will return balls

Bodddd:


liliia-ponomarenko (changes requested):
Let’s improve your solution😉

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
liliia-ponomarenko, line 10:
```suggestion
        return Color.values()[index].toString();
```

liliia-ponomarenko (approved):
great 😎

PR #500, opened 2022-09-27
implemented method getRandomBall

nazariipertsak:


liliia-ponomarenko (changes requested):
Let’s improve your solution😉

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
```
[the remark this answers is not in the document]
nazariipertsak: if i understand right, if i create constructor, then i need to create an instance of Lottery in method getRandomBall ?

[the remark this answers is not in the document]
nazariipertsak: > it's bad practice to initialize dependencies in the method. you can initialize it in fields,

if i understand right, if i create constructor, then i need to create an instance of Lottery in method getRandomBall ?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,20 @@
...
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+        Lottery lottery = new Lottery(colorSupplier, random);
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
```
liliia-ponomarenko, line 18: Can we use only one loop?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public Lottery(ColorSupplier colorSupplier, Random random) {
+        this.colorSupplier = colorSupplier;
+        this.random = random;
+    }
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
liliia-ponomarenko, line 17:
```suggestion
        ball.setNumber(random.nextInt(100));
```
magic number

liliia-ponomarenko (approved):
good job! 👍

src/main/java/core/basesyntax/Lottery.java
nazariipertsak: > magic number

Thank you! Already edited

PR #501, opened 2022-09-27
Added the ability to create random balls

romangolovatyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
hronograf, line 9: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
```
hronograf, line 6: it is a good practice to create already valid object
so it would be better to add constructor with all fields to our class)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Lottery {
+    private static final int BALL_COUNT = 101;
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(BALL_COUNT));
+        return ball;
+    }
```
hronograf, line 13: We can add fields with type `Random` and `ColorSupplier` to our class, so we won't need to create instances every time

hronograf (changes requested):
Good job!

src/main/java/core/basesyntax/Lottery.java
romangolovatyi: Thank you for spending your time and giving advice

hronograf (approved):
Great! 🚀

PR #502, opened 2022-09-27
Hw solve Anton

CyberDogFK:


liliia-ponomarenko (changes requested):
Let’s improve your solution😉

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class Application {
+    /* number of showing random balls */
+    public static final int NUMBER_OF_BALLS = 3;
```
liliia-ponomarenko, line 7:
```suggestion
    private static final int NUMBER_OF_BALLS = 3;
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public String getRandomColor(Random random) {
+        // random integer in length of number of BallColor
+        // values
```
liliia-ponomarenko, line 8:
```suggestion
```

src/main/java/core/basesyntax/BallColors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum BallColors {
```
liliia-ponomarenko, line 3:
```suggestion
public enum BallColor {

```
or just Color

liliia-ponomarenko (approved):
good job! 👍

PR #507, opened 2022-09-27
Lottery homework

anastasiavasylyk:


liliia-ponomarenko (changes requested):
Let’s improve your solution 😉

liliia-ponomarenko (approved):
great🔥

PR #509, opened 2022-09-27
Implemented method getRandomBall

denys-domashevskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
+        }
```
hronograf, line 7: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)
[All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
hronograf, line 7: We can add field with type Random to our class, so we won't need to create an instance every time when method invoked

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[index].toString();
```
hronograf, line 8: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor().toString(),
+                new Random().nextInt(100));
+    }
```
hronograf, line 9: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
We can add fields with type Random and ColorSupplier to our class, so we won't need to create instances every time

hronograf (changes requested):
Good!

PR #513, opened 2022-09-27
Classes Ball, Lottery and enum Colors created. Methods getRandomColor…

luka-lozovyk:
…() and getRandomBall() implemented.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Locale;
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].toString().toLowerCase(Locale.ROOT);
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
     }
```
hronograf, line 11: Why do we need here few instances of type `Lottery`?
Also, it would be better to use a cycle for repeatable code

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Locale;
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
hronograf, line 8: We can add field with type `Random` to our class, so we won't need to create an instance every time when method invoked

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        return Colors.values()[index].toString().toLowerCase(Locale.ROOT);
```
hronograf, line 9: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
+    }
```
hronograf, line 8: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
also, we can add fields with type `Random` and `ColorSupplier` to our class, so we won't need to create instances every time

hronograf (changes requested):
Nice!

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
...
 
+import static core.basesyntax.Lottery.getRandomBall;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = getRandomBall();
+            System.out.println(ball);
+        }
```
hronograf, line 10: do we need `ball` variable here?

hronograf (approved):
Good job! 🔥

PR #514, opened 2022-09-28
A task is released to create balls with random color and number.

OleksandrKorniychuk:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
hronograf, line 9: what if we want to print 10 balls? or 100?
we don't want to repeat our code (we can use cycle)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
hronograf, line 7: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
also, we can add fields with type `Random` and `ColorSupplier` to our class, so we won't need to create instances every time

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public Ball(Colors color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
```
hronograf, line 15: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)
(in this case `toString` will be invoked automatically for `color`)

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        final ColorSupplier colorSupplier = new ColorSupplier();
+        final int random = new Random().nextInt(100);
+        return new Ball(colorSupplier.getRandomColor(), random);
+    }
```
hronograf, line 10: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
read more carefully about magic numbers
also, let's add fields with type `Random` and `ColorSupplier` to our class, 
as you did it for `Random` in `ColorSupplier` class

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public String getBallInfo() {
+        return new StringBuilder()
+                .append("Ball color = ")
+                .append(color)
+                .append(", ")
+                .append("number = ")
+                .append(number)
+                .toString();
+    }
```
hronograf, line 20: it would be better to override `toString` method instead of creating this one (because `toString` is a standard method for string representation of an object)

also, it is better to use `color.name()` 
(`.append(color)` will call `toString`⬇️ )

hronograf (changes requested):
Nice!
also, it is better to write more informative commit messages 
(like `fixed comments` and not like `It is difficult to understand...`)

tasks can be difficult, but you should study hard to become a good programmer)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
...
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color=" + color.name()
+                + ", number=" + number
+                + '}';
+    }
```
hronograf, line 18: nice!

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        int times = 3;
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < times; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
hronograf, line 9: can we write just `System.out.println(lottery.getRandomBall());`?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final int random = new Random().nextInt(100);
```
hronograf, line 7: now i think all balls were created with the same number
field type should be `Random`, not `int`

also
[All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
number 100 is a magic number.
it isn't obvious what does it mean, so it is better to give an informative name for this number

hronograf (changes requested):
Better and better!

src/main/java/core/basesyntax/Application.java
OleksandrKorniychuk: Of course!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int range = 100;
```
hronograf, line 6: constant means `private static final` (sometimes `public static final`)
[All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)

hronograf (approved):
Great! 🏎️ 
and @OleksandrKorniychuk, please, check the comment)

PR #516, opened 2022-09-28
Implemented Lottery class with ColorSupplier subclass with generator methods, Ball class, Colors enum

muzykaVP:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
hronograf, line 8: Can we write just `System.out.println(lottery.getRandomBall());`?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random randomColor = new Random();
```
also, let's find a better name for `randomColor` variable, because it isn't a color) 
(just `random` will be ok)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
```
hronograf, line 5: Do we need inheritance here? (`Lottery extends ColorSupplier`)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
...
+    public Ball getRandomBall() {
+        Random randomNumber = new Random();
+        return new Ball(getRandomColor(), randomNumber.nextInt(100));
+    }
```
hronograf, line 9: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
We can add field with type `Random` to our class, so we won't need to create an instance every time when method invoked

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
hronograf, line 7: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants) (I am about number `3`)

hronograf (changes requested):
Good job!

src/main/java/core/basesyntax/Application.java
muzykaVP: I just cheked, it really did write same result. I tested what would be  printed without overrided toString() method and got printed just references of  objects. I got the idea that if classes have toString() method it will be called. Correct me if  i'm wrong.

src/main/java/core/basesyntax/ColorSupplier.java
muzykaVP: Got it

src/main/java/core/basesyntax/Lottery.java
muzykaVP: My bad

hronograf (changes requested):
Nice!
let's remove `extends ColorSupplier` from `Lottery` class and fix other comments)

src/main/java/core/basesyntax/Lottery.java
muzykaVP: If there exists an IS-A relationship between two classes, like for example Car IS-A Vehicle.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int MAGIC_THREE = 3;
```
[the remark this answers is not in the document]
muzykaVP: : )

hronograf (approved):
Cool! 😎

PR #517, opened 2022-09-28
create classes Ball, Lottery; create Enum Colors; implemented…

yehbreus:
…getRandomColor and getRandomBall methods; added constructor to Ball class, and override toString method; added psvm

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,10 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
+
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return new String(color + " " + number);
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int rand = new Random().nextInt(Colors.values().length);
+        return Colors.values()[rand].toString();
     }
 }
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return new String(color + " " + number);
```
hronograf, line 14: do we need `new String` here?
also, we can make `toString` more informative (like "Ball{color=.., number=...}")

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int rand = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[rand].toString();
```
hronograf, line 8: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
hronograf, line 6: after we add field with type `Random`, we should remove `static` modifier to use that field inside our method

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
hronograf, line 6: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_RANDOM_NUMBER = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), new Random()
+                .nextInt(MAX_RANDOM_NUMBER + 1));
+    }
```
hronograf, line 12: here we have same problem as we had in `ColorSupplier` class: 
we create new instance of `Random` class for every method run

hronograf (changes requested):
almost done!

hronograf (approved):
Good job! 💎

yehbreus:
> Good job! 💎

ty)

PR #518, opened 2022-09-28
created classes Ball, Lottery. Implemented method getRandomColor()

Danyl-Kniaziev:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return BallColor.values()[new Random().nextInt(BallColor.values().length)].name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
hronograf, line 7: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
We can add fields with type `Random` and `ColorSupplier` to our class, so we won't need to create instances every time

hronograf (changes requested):
Good job!

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
 public class Application {
-    public static void main(String[] args) {
+    public void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[] {
+                lottery.getRandomBall(),
+                lottery.getRandomBall(),
+                lottery.getRandomBall()
+        };
```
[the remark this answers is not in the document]
Danyl-Kniaziev: > we can use cycle to fill our array
is my method of filling the array bad?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
 public class Application {
-    public static void main(String[] args) {
+    public void main(String[] args) {
```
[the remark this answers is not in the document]
Danyl-Kniaziev: it was an accident, I wrote a static methode, then I corrected it when I saw the typical mistakes, and then I corrected the main :sweat_smile:

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+
+    }
```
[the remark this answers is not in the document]
Danyl-Kniaziev: ok. wanted the best

src/main/java/core/basesyntax/ColorSupplier.java
Danyl-Kniaziev: I understand that the code becomes less understandable because of such constructions. But in typical errors it was said -  don`t create additional variables, if possible. Where is this fine line?

src/main/java/core/basesyntax/Lottery.java
Danyl-Kniaziev: Similar question 👆

src/main/java/core/basesyntax/ColorSupplier.java
Danyl-Kniaziev: Or is the problem that we will create, each time a new object?

src/main/java/core/basesyntax/Lottery.java
Danyl-Kniaziev: I don't understand where the constants are

src/main/java/core/basesyntax/Application.java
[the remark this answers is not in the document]
hronograf: for this task it isn't bad, and imagine that we should add 100 or 1000 balls to array
so let's write code that can do such tasks easy

src/main/java/core/basesyntax/ColorSupplier.java
hronograf: yes, the problem is that we create a new object for each method call

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
hronograf, line 10: [All magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)
number 100 is a magic number.
it isn't obvious what does it mean, so it is better to give an informative name for this number

hronograf (changes requested):
let's do one small fix)

hronograf (approved):
Great! 🔥

PR #529, opened 2022-09-28
creating 3 balls and printing info about them

AlinaOryshak:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+        for (int i = 0; i < balls.length; i++) {
+            System.out.println(balls[i].toString());
```
hronograf, line 11: can we write just `System.out.println(balls[i]);` here?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball() {
+    }
```
hronograf, line 8: do we need this constructor?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball number " + this.getNumber() + " has " + this.getColor() + " color.";
```
hronograf, line 33: if we concatenate some object with a string, `toString` method will be called, 
but we should avoid this for enum constants

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,13 @@
...
+    red,
+    orange,
+    yellow,
+    green,
+    blue,
+    violet,
+    rose,
+    black,
+    white
+}
```
hronograf, line 13: read about [Constant names](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

hronograf (changes requested):
Cool!
let's improve this solution a bit)

hronograf (approved):
Nice! 💎

PR #535, opened 2022-09-29
Hw loterry

Mar4uk21:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery ball = new Lottery();
+        System.out.println(ball.getRandomBall());
+        Lottery ball1 = new Lottery();
+        System.out.println(ball1.getRandomBall());
+        Lottery ball2 = new Lottery();
+        System.out.println(ball2.getRandomBall());
     }
```
hronograf, line 11: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)
let's use cycle here, so we can easily print 3, 10 or 100 balls
also, `ball` is not the best name for variable with type `Lottery`)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Ball extends ColorSupplier {
```
hronograf, line 3: do we need inheritance here?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,11 @@
...
+    private String color;
+    private int number;
+
+    @Override
+    public String toString() {
+        return super.toString();
```
hronograf, line 9: what did you see in your console when you printed this balls?
let's change this implementation) 
(we should be able to see that it is a ball, its color and number)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+public class Lottery extends Ball {
+    private static final int FIGURE_COUNT = 100;
+    private Random random = new Random();
+
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        int number = random.nextInt(FIGURE_COUNT);
+        String color = String.valueOf(ball.getRandomColor());
+        return color + number;
+    }
```
hronograf, line 14: `getRandomBall` method should return instance of `Ball`, not `String` (because of its name)
[Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants) (`String.valueOf` isn't so good)

also it would be better to have an appropriate constructor in the `Ball` class, 
so we could write something like `new Ball(color, number);`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
```
hronograf, line 5: do we need inheritance here?
please, check in which cases we need inheritance

hronograf (changes requested):
Nice!

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        Ball ball = new Ball();
```
hronograf, line 9: do we need ball here?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int FIGURE_COUNT = 100;
```
hronograf, line 6: it would be better to find better name for this constant) (like `MAX_BALL_NUMBER`)

hronograf (changes requested):
Good!

hronograf (approved):
Great! 💥

PR #537, opened 2022-09-29
hw-lottery

Iryna-a:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
+        }
+
```
hronograf, line 13: [Please don’t add redundant empty lines to your code.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#please-dont-add-redundant-empty-lines-to-your-code)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private static final int BOUND = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Ball ball = new Ball();
+
+    public Ball getRandomBall() {
+        ball.setNumber(random.nextInt(BOUND));
+        ball.setColor(colorSupplier.getRandomColor());
+        return ball;
+    }
```
hronograf, line 15: we should create new ball each time `getRandomBall` was called (but now we just change the same object)

also it would be better to have a constructor for `Ball` class (like `public Ball(String color, int number)`), 
so we could create our balls with parameters that we need (like `new Ball(someColor, someNumber)`)

hronograf (changes requested):
Good job!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Lottery {
+    private static final int BOUND = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(colorSupplier.getRandomColor(), random.nextInt(BOUND));
+        return ball;
+    }
```
hronograf, line 13: [Don’t create temporary variables if you can directly return value from the method](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-temporary-variables-if-you-can-directly-return-value-from-the-method)

hronograf (approved):
Nice! 💎 
and @Iryna-a, please, check the comment)

PR #538, opened 2022-09-29
added enum and 2 classes with override in ball

Alex-Kharchuk:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball {
+    private Color color = ColorSupplier.getRandomColor();
+    private int number = new Random().nextInt(100);
```
hronograf, line 7: what if we want to create a ball with some specific color and number?
(it would be better to have a constructor for this fields)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,29 @@
...
+        this.number = number;
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball color: " + this.color + ". Ball number: " + this.number;
```
hronograf, line 27: when we write `someObject + someString`, then `someObject.toString()` will be invoked automatically

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Color {
+    black,
+    blue,
+    red,
+    yellow,
+    green,
+    white
+}
```
hronograf, line 10: please, read [Constant names](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
+        int colorIndex = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball1 = new Ball();
+        return ball1;
+    }
```
hronograf, line 7: [Don’t create temporary variables if you can directly return value from the method](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-temporary-variables-if-you-can-directly-return-value-from-the-method)

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        int numberOfBalls = 3;
+        Ball[] lotteryBalls = new Ball[numberOfBalls];
+        for (int i = 0; i < numberOfBalls; i++) {
+            Lottery randomBall = new Lottery();
+            lotteryBalls[i] = randomBall.getRandomBall();
+            System.out.println(lotteryBalls[i].toString());
```
hronograf, line 11: do we need `toString()` here?

hronograf (changes requested):
Good!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball();
```
KostyaNikiforov, line 5: This method must contains a real implementation. But it returns always an empty Ball object. So why do we need this class at all?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball {
+    private Color color = ColorSupplier.getRandomColor();
+    private int number = new Random().nextInt(100);
```
KostyaNikiforov, line 7: This code must be in the Lottery class. 
He is responsible for creating Ball objects.

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #540, opened 2022-09-29
created and implemented getRandomBall method in the Lottery class

ana-rom:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Ball {
+    public String color;
+    public int number;
```
liliia-ponomarenko, line 5: Let’s make these fields private. Follow to the principle of encapsulation

PR #541, opened 2022-09-29
created and implemented getRandomBall method in the Lottery class

ana-rom:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
...
+import java.util.Arrays;
+
 public class Application {
+    private static final int AMOUNT_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[AMOUNT_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
liliia-ponomarenko, line 12: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,12 @@
...
+
+public enum Color {
+    BLACK,
+    BLUE,
+    GREEN,
+    ORANGE,
+    PINK,
+    RED,
+    VIOLET,
+    YELLOW;
```
liliia-ponomarenko, line 11:
```suggestion
    YELLOW
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BALL_COUNT = 100;
+
+    public Ball getRandomBall() {
+        Ball randomBall = new Ball();
+        randomBall.setColor(new ColorSupplier().getRandomColor());
+        randomBall.setNumber(new Random().nextInt(BALL_COUNT));
```
liliia-ponomarenko, line 11: it's bad practice to initialize dependencies in the method. you can initialize it in fields, 
 but good practice initialization using a constructor:
```
    private final Random random;
    private final ColorSupplier supplier;

    public Lottery(Random random, ColorSupplier supplier) {
        this.random = random;
        this.supplier = supplier;
    }
```

liliia-ponomarenko (changes requested):
Let’s fix a few mistakes😉

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
```
liliia-ponomarenko, line 6:
```suggestion
```

liliia-ponomarenko (approved):
LGTM🔥

PR #542, opened 2022-09-29
getRandomBall() method implemented

Sachamunn:


liliia-ponomarenko (changes requested):
Let’s  check ALL common mistakes  again😉

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final int COLOR_COUNT = 5;
     public String getRandomColor() {
-        return null;
+        return Color.values()[COLOR_COUNT].name();
```
ChumakAn, line 8: you should use dynamic number instead of COLOR_COUNT which is always 5, you need to generate the number first randomly

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
ChumakAn, line 7: 100 - is a magic number, should be a constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        System.out.println(new Lottery().getRandomBall().toString());
```
ChumakAn, line 5: It's better to use loop, cause now you are repeating yourself

ChumakAn (changes requested):
Let's make your code better)

ChumakAn (approved):
LGTM

PR #550, opened 2022-10-05
released lottery

rozyzlav:


src/main/java/core/basesyntax/DiffColor.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum DiffColor {
+    blue,
```
ChumakAn, line 4: it's a good practice to write enum values with upper case, like **BLUE , not blue**

ChumakAn (approved):
Well done! One comment for better code style👍

PR #554, opened 2022-10-10
NewCommit002-done1-jv_lottery

Poikka001:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static DifferentColor getRandomColor() {
```
zachar-chorny, line 6: Methods don't have to be static, you have to create class objects to use their methods

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        int index = new Random().nextInt(DifferentColor.values().length);
+        return DifferentColor.values()[index];
```
zachar-chorny, line 8:
```suggestion
    private final Random random = new Random();
    
    public static DifferentColor getRandomColor() {
        int index = random.nextInt(DifferentColor.values().length);
        return DifferentColor.values()[index];
```
It is better to make Random a class field so that it is not created every time a method is called

src/main/java/core/basesyntax/DifferentColor.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum DifferentColor {
```
zachar-chorny, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+
+    public static int randome() {
+        Random myRandom = new Random();
+        return myRandom.nextInt(100);
+    }
+
+    public static Ball getRandomBall() {
```
zachar-chorny, line 12: The method must not be static. Just create an object of Lotery class in main method

zachar-chorny (changes requested):

kateryna-deshchinska (approved):

PR #555, opened 2022-10-10
added solution for jv-lottery

vgladkyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
zachar-chorny, line 7: You create an object of the Lottery class in each iteration of the loop. Put it in a variable before the loop and use it in the loop

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public enum Color {
+    Red,
```
zachar-chorny, line 4: In enum, all entries must be in the aper case

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
zachar-chorny, line 7:
```suggestion
    private final Random random = new Random();

    public Color getRandomColor() {
        int index = random.nextInt(Color.values().length);
```
In this case, an object of the Random class will not be created each time the getRandomColor() method is called. It will be created only once

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
```
zachar-chorny, line 9: Do this with Random and ColorSupplier as I showed above

zachar-chorny (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BALL_NUMBER = 100;
+    private final ColorSupplier randomColor = new ColorSupplier();
```
Rommelua, line 7:
```suggestion
    private final ColorSupplier сolorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,17 @@
...
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return color + " " + number;
+    }
+
```
Rommelua, line 16: empty line

Rommelua (changes requested):
Good work, just two comments

src/main/java/core/basesyntax/Lottery.java
vgladkyi: thank you!

kateryna-deshchinska (approved):

PR #566, opened 2022-10-18
Lottery homework solution.

Stim1:
Created new reposytory with all fixed mistakes .

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+
+    public static final int BALL_NUMBER = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_NUMBER = 100;
```
aaananas, line 6:
```suggestion
    private static final int MAX_NUMBER = 100;
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+
+    public static final int BALL_NUMBER = 3;
```
aaananas, line 5:
```suggestion
    private static final int BALL_NUMBER = 3;
```

aaananas (approved):

PR #570, opened 2022-10-22
jv-Lottery from Dmytro Koval 

dimonkoval:
jv-Lottery from Dmytro Koval


aaananas (changes requested):
image

run `mvn clean package` and fix errors

PR #611, opened 2022-11-16
create class Lottery with random methods

InnaDiachkina:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
Nick97-git, line 7: create a global field Random and use it here, so you won't need to create its instance each time this method is called

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
Nick97-git, line 11: create a global fields Random and ColorSupplier and use them here, so you won't need to create their instances each time this method is called

Nick97-git, line 11: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int BALL_NUMBER = 100;
```
```suggestion
    private static final int MAX_BALL_RANDOM NUMBER = 100;
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BALL_NUMBER = 100;
```
```suggestion
    private static final int MAX_BALL_RANDOM_NUMBER = 100;
```

Nick97-git (changes requested):

Nick97-git (approved):

PR #635, opened 2022-12-05
Hw 03 solution

kopervas:


kopervas (commented):
Done randomColor a class-level variable.

aaananas (approved):

PR #645, opened 2022-12-07
Three random balls were created and printed in console

taras-zhyravetskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery selectedBall = new Lottery();
```
Nick97-git, line 5: does it look like a ball? selectedBall -> lottery

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + this.color + ", Number: " + this.number;
```
Nick97-git, line 25: don't need to use this.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
Nick97-git, line 7: make a global Random field and use it here, so you won't need to create each time its instance when you call this method

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private Random random = new Random();
+    private Ball ball = new Ball();
```
Nick97-git, line 8: make this variable local

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
```
```suggestion
    private static final int MAX_RANDOM_BALL_NUMBER = 100;
```

Nick97-git (changes requested):

src/main/java/core/basesyntax/Ball.java
taras-zhyravetskyi: Don't understand. In task we should override toString() method.

src/main/java/core/basesyntax/Lottery.java
taras-zhyravetskyi: In this case, we won't to create each time its instance when call this method?

Nick97-git (approved):

PR #656, opened 2022-12-08
Wrong code for fixing mistakes

Butovych:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Object getRandomBall() {
```
```suggestion
    public Ball getRandomBall() {
```
you should create an instance of Lottery in Main to use its method

Nick97-git (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,21 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANDOM_BALL_NUMBER_BOUND = 100;
```
```suggestion
    private static final int RANDOM_BALL_NUMBER_UPPER_BOUND = 100;
```

Nick97-git (approved):

PR #675, opened 2022-12-21
jcoop lottery solution

imhatep:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall().toString());
         // create three balls using class Lottery and print information about them in console
```
karma-o, line 7: it should be 3 balls, not 1

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
karma-o, line 7: you are creating new `Random` instance each time you are calling this method, we should avoid that

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
karma-o, line 7: same as `Random` move it to a class-level variable

karma-o (changes requested):
you also changed some other files that are not specified by the task, revert it please

PR #690, opened 2022-12-29
create class ball with related methods

ArtyomZayarny:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random randomColor = new Random();
+        int index = randomColor.nextInt(4);
```
karma-o, line 8: Lets use `Colors.values().length()` instead of hardcoding 4 here

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public void setColor() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        this.color = colorSupplier.getRandomColor();
+    }
+
+    public void setNumber() {
+        Random random = new Random();
+        this.number = random.nextInt(100);
+    }
```
karma-o, line 17: setters should not have any logic inside them, they should only set a field of our object that it gets in the parameters, `Lottery` class should be responsible for creating a ball with random parameters by using our `ColorSupplier` class in a [composition-manner](https://www.infoworld.com/article/3409071/java-challenger-7-debugging-java-inheritance.html#:~:text=Inheritance%20and%20composition%20are%20two,the%20sum%20of%20its%20parts.) structure

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+
+    @Override
+    public String toString() {
+        return this.getClass().getSimpleName()
+                + "color: "
+                + this.color
+                + ", number: "
+                + this.number;
+    }
```
karma-o, line 26: we need `this` keyword only for when we have similar variable naming, or we want to reference a current object
```suggestion
        return this.getClass().getSimpleName()
                + " color: " + color
                + ", number: " + number;
    }
```

karma-o (changes requested):

src/main/java/core/basesyntax/Ball.java
karma-o: also, don't forget about making magic numbers (like this `100`) a constants with a proper naming

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+        System.out.println(
+                "ball1: " + ball1 + "\n"
+                 + "ball2: " + ball2 + "\n"
+                 + "ball3: " + ball3
```
aaananas, line 12: image

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
```suggestion
public class ColorSupplier {
    private final Random random = new Random();
    
    public String getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
-        return null;
+        Random randomColor = new Random();
+        int index = randomColor.nextInt(Colors.values().length);
```
```suggestion
        int index = random.nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+        Colors[] values = Colors.values();
+        return values[index].toString();
```
aaananas, line 10: image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
classes are usually called in singular form

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(100);
+        ball.setColor();
```
```suggestion
        ball.setColor(colorSupplier.getRandomColor());
```
you should call ColorSuppliers method here to set a color

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(100);
```
aaananas, line 7:
```suggestion
        ball.setNumber(MAX_NUMBER);
```
avoid magic numbers - create a constant (private static final)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public void setColor() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        this.color = colorSupplier.getRandomColor();
+    }
+
+    public void setNumber(int num) {
+        Random random = new Random();
+        this.number = random.nextInt(num);
+    }
```
aaananas, line 17:
```suggestion
    public void setNumber(int num) {
        this.number = num;
    }
```
you should get a number here and just set it

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Random;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public void setColor() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        this.color = colorSupplier.getRandomColor();
```
aaananas, line 11:
```suggestion
    public void setColor(String color) {
        this.color = color;
```

aaananas (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
...
+        this.color = color;
+    }
+
+    public void setNumber(int num) {
+        this.number = num;
+    }
+
+    @Override
+    public String toString() {
+        return "color: " + this.color + ", number: " + this.number;
```
Rommelua, line 17:
```suggestion
        return "color: " + color + ", number: " + number;
```

Rommelua (approved):

PR #693, opened 2022-12-30
feat: jv-lottery #comment lottery assignment was implemented

sania517:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return (new StringBuilder("{color: "))
+                .append(color)
+                .append("; number: ")
+                .append(number)
+                .append("}")
+                .toString();
```
okuzan, line 19: I don't think we gain something here using StringBuilder

okuzan (changes requested):

okuzan (approved):
Good job)

PR #695, opened 2022-12-31
Solution1

Dominskyi93:


src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
sarakhmen, line 6: It's better not to create several instances of the same class if you can use only one instance for your purposes

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.security.SecureRandom;
+
+public class Lottery {
+    private final int randomNumber = new SecureRandom().nextInt(100);
```
sarakhmen, line 6: The randomNumber of instances of this class is hardcoded, so it's impossible to reuse them. It's better to get a random number on each getRandomBall() call.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.security.SecureRandom;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final int randomNumber = new SecureRandom().nextInt(Color.values().length);
```
sarakhmen, line 6: The randomNumber of instances of this class is hardcoded, so it's impossible to reuse them. It's better to get a random number on each getRandomColor() call.

sarakhmen (changes requested):
Good solution! Let’s improve it 😉

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.security.SecureRandom;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
```
Andriy-dz, line 6: let's return a String, not a Color

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
...
+        return Color.values()[new SecureRandom().nextInt(Color.values().length)];
```
Andriy-dz, line 7: let's use the Random class instead of SecureRandom and make it a constant so that we don't have to create it every time

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.security.SecureRandom;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new SecureRandom().nextInt(100));
```
Andriy-dz, line 7: magic number. Lets use the Random class instead of SecureRandom and make it a constants to ColorSupplier and Random

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Andriy-dz, line 6: magic number

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        int numberOfBalls = 3;
```
Andriy-dz, line 5: lets make constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANDOM_NUMBER = new Random().nextInt(Color.values().length);
```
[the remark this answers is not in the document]
Dominskyi93: But it is constant, I don't understand

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final int numberOfColor = new Random().nextInt(Color.values().length);
```
KseniiaMakarova, line 6: the problem is that when you create `new ColorSupplier()` - this variable is initialized and is no further changed
so your number is not randomly generated each time you call the `getRandomColor()` method on the same ColorSupplier object - and this is a mistake

please move this variable to the method - so it could be reset each time you call the method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int randomNumber = new Random().nextInt(100);
```
KseniiaMakarova, line 6: same here!
move generation of random number to the method

but 100 is a magic number and it should be a constant field (final static)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Lottery().randomNumber);
```
KseniiaMakarova, line 9: you don't need to create a Lottery object, you can use the field `randomNumber` directly

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Main {
+    private static final int NUMBER_OF_BALLS = 3;
+
+    public static void main(String[] args) {
+
```
KseniiaMakarova, line 7: redundant empty line

KseniiaMakarova (changes requested):
please next time be attentive to the branch where you work
you should NOT work in master branch and make a PR from master

src/main/java/core/basesyntax/Main.java
Dominskyi93: "main" should be separeted from previous statement!

KseniiaMakarova (changes requested):
something is wrong with your code - check fails
let's find the mistake

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
```
KseniiaMakarova, line 6: redundant empty line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
...
+    public int getRandomNumber() {
+        Random random = new Random();
+        return random.nextInt(Color.values().length);
+    }
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[this.getRandomNumber()].name();
```
KseniiaMakarova, line 13:
```suggestion
        return Color.values()[getRandomNumber()].name();
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #696, opened 2022-12-31
Lottery homework

Dominskyi93:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                +
+                "color='" + color + '\''
+                +
+                ", number=" + number
+                +
+                '}';
```
petro-ushchuk, line 31: Remove extra line breaks before concatenation.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int i = new Random().nextInt(Color.values().length);
```
petro-ushchuk, line 7: Don't create Random instance every time you call getRandomColor, better to make it constant.
Also, better to use SecureRandom instead.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
+        return ball;
```
petro-ushchuk, line 10: Create constructor with params and inline return instance directly.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
petro-ushchuk, line 9: Make Random and ColorSupplier consant variables of class.
Also, better to use SecureRandom implementation.

petro-ushchuk, line 9: Make this magic number consntant.

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
```
petro-ushchuk, line 7: Don’t create several instances of the Lotery if you can use only one instance for your purposes

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Color {
+    RED, GREEN, BLUE, YELLOW, BLACK, WHITE, PINK, BROWN, GREY, PURPLE
```
karma-o, line 4: maybe not a mistake but in my opinion it is better to put each on the new line

karma-o (changes requested):

PR #697, opened 2023-01-02
- Created Ball class with the following fields: color and number and …

Vladyslav-Voronkov:
…override toString() method

- Created enum with different colors
- Implemented method getRandomColor() in class ColorSupplier. This method returning a random color, based on the enum values (used Random class for this)
- Created a Lottery class with getRandomBall() method, which returned the ball. This method created a ball with a random color and a random number (the maximum possible number is 100) and returned this ball from the method.

To test the program, in main() method, created three balls using getRandomBall() method from Lottery class. And printed information about them in the console.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lt1 = new Lottery();
+        Lottery lt2 = new Lottery();
+        Lottery lt3 = new Lottery();
+        System.out.println(lt1.getRandomBall());
+        System.out.println(lt2.getRandomBall());
+        System.out.println(lt3.getRandomBall());
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
+        int randIndex = random.nextInt(Colors.values().length);
+        return Colors.values()[randIndex].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
+    WHITE,
+    BLACK,
+    GREEN,
+    RED,
+    PURPLE,
+    ORANGE
+}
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

okuzan (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private ColorSupplier supplier = new ColorSupplier();
+    private Random random = new Random();
+    private int maxRand = 100;
```
okuzan, line 9: make this constant  (static final + naming)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
```
okuzan, line 6: delete empty line

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+    private ColorSupplier supplier = new ColorSupplier();
+    private Random random = new Random();
+    private int maxRand = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(supplier.getRandomColor(),random.nextInt(maxRand));
```
okuzan, line 12:
```suggestion
        return new Ball(supplier.getRandomColor(), random.nextInt(maxRand));
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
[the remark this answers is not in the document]
okuzan: not fixed, do it the same way as in `Lottery`

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lt1 = new Lottery();
+        Lottery lt2 = new Lottery();
+        Lottery lt3 = new Lottery();
+        Lottery[] lotteries = new Lottery[] { lt1,lt2,lt3 };
+        for (Lottery lottery : lotteries) {
+            System.out.println(lottery.getRandomBall());
+        }
```
okuzan, line 11: You just need one lottery instance, use for-loop (instead of for-each) with bound 3 if you want to run test three times. Then in for loop just call `getRandomBall`. It's designed to return random ball after all for the same one instance :)

okuzan (changes requested):

okuzan (approved):

PR #707, opened 2023-01-06
task completed

MaksymHorielikov:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].name();
```
petro-ushchuk, line 7: Don't create Random instance every time you call getRandomColor, better to make it constant.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(),
+                new Random().nextInt(MAX_NUMBER + 1));
```
petro-ushchuk, line 10: Make Random and ColorSupplier consant variables of class.
Also, better to use SecureRandom implementation.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
+    BLACK, WHITE, BLUE, GREEN, RED, ORANGE, YELLOW, BROWN
```
karma-o, line 4: in my opinion, having them split by a new line would be better

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private static final Random RANDOM = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(),
```
karma-o, line 10: we are still creating a new instance of `ColorSupplier` each time we call this method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+                RANDOM.nextInt(MAX_NUMBER + 1));
```
karma-o, line 11: just curious why `+ 1`?

karma-o (changes requested):

src/main/java/core/basesyntax/Lottery.java
MaksymHorielikov: nextInt(int bound) method returns:
the next pseudo-random, uniformly distributed int value between zero (inclusive) and boundary (EXCLUSIVE) from this random number generator's sequence

MaksymHorielikov: i add public getRandomNumberFromRange, returns random int in range [firstInteger, secondInteger] which inclusive this two integer

sarakhmen (approved):
Good job) 🔥

PR #714, opened 2023-01-12
Develop

MaxSkory:
I changed the getRandomColor() method modifier to static and changed the returned data type of this method to Color. Think, it will be better, in this case we don't need to create a new instance of ColorSupplier and call toString() method with each Enum.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,9 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall().toString());
+        System.out.println(lottery.getRandomBall().toString());
+        System.out.println(lottery.getRandomBall().toString());
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,7 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
+        // create three balls using class Lottery and print information about them in console
         Lottery lottery = new Lottery();
         System.out.println(lottery.getRandomBall().toString());
         System.out.println(lottery.getRandomBall().toString());
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -11,9 +11,6 @@ public class Ball {
 
     @Override
     public String toString() {
-        return "Ball{" +
-                "color=" + color +
-                ", number=" + number +
-                '}';
+        return "Ball{" + "color=" + color + ", number=" + number + '}';
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall().toString());
```
nacenik, line 7: Use a loop to print balls. Don't use method `toString()`, cause `System.out.println()` call it by default

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int number;
+
+    public Ball(Color color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{" + "color=" + color + ", number=" + number + '}';
```
nacenik, line 14: [Use name() for getting String representation of enum](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
nacenik, line 7: Random should be class-level field.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(100));
```
nacenik, line 7: Random and  ColorSupplier should be class-level fields.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
```
nacenik, line 6: [getRandomColor() should be non-static](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-use-static-methods-in-your-solution)

src/main/java/core/basesyntax/Lottery.java
nacenik: All magic number should be constant

nacenik (changes requested):

okuzan (approved):
Good job!

PR #715, opened 2023-01-12
classes were created, methods getRandomColor() and getRandomBall() we…

oleksiistepaniak:
…re implemented

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String toString() {
+        return "The color of the ball is " + this.color + " and the number of the ball is "
+                + this.number;
+    }
+
+    // setters
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String toString() {
```
maxline, line 7: please read about ordering of the methods.
`toString()` should be moved to the end of the class.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
maxline, line 7: Do we really need to create a new instance of `Random` every time when this method is called?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        Ball ball = new Ball();
+        int randomNumber = new Random().nextInt(100);
```
It is better to create a new fields as class properties.

maxline (changes requested):
please take a look into comments above

oleksiistepaniak:
> 

I want to thank you for your remarks. I understand that I need to learn how to write readable and clean code. I've fixed everything that I have to fix.

karma-o (approved):

PR #718, opened 2023-01-20
created several classes and enum, implemented two method and printed …

knyazkyylubomir:
…information about balls

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] balls = new Lottery[NUMBER_OF_BALLS];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery();
+            System.out.println(balls[i].getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString();
+    }
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier supplier = new ColorSupplier();
+
+    public String getRandomBall() {
+        StringBuilder builder = new StringBuilder();
+
+        ball.setColor(supplier.getRandomColor());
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] balls = new Lottery[NUMBER_OF_BALLS];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery();
+            System.out.println(balls[i].getRandomBall());
+        }
```
okuzan, line 12: how about this: let's use NUMBER_OF_BALLS constant in for loop condition directly, and don;t create an array at all, just call new Lotter().getRandomBall()

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString();
```
okuzan, line 25: you need to manually write the logic for toString , using string concatenation, don't forrget to include color, number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+public class Lottery {
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier supplier = new ColorSupplier();
+
+    public String getRandomBall() {
+        StringBuilder builder = new StringBuilder();
+
+        ball.setColor(supplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
okuzan, line 14: 100 should be class constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier supplier = new ColorSupplier();
+
+    public String getRandomBall() {
```
okuzan, line 10: this method should return Ball

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
...
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery balls = new Lottery();
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            System.out.println(new Lottery().getRandomBall());
+        }
```
Ivan95kos, line 10: Try not to create variables that you don't use.
```suggestion
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(new Lottery().getRandomBall());
        }
```
Or if you have already created it, you can use the following
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return this.color + " " + this.number;
```
Ivan95kos, line 25: In the future, try to write a better description

Ivan95kos (approved):

src/main/java/core/basesyntax/Ball.java
knyazkyylubomir: Ok, I'll pay more attention next time

PR #723, opened 2023-01-28
hw solution

KyrylH:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color = null;
+
+    private int number = 0;
```
petro-ushchuk, line 6: It is not necessary to set default values, because they are already set.

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Color {
+    White,
+    Blue,
+    Pink,
+    Green,
+    Yellow,
+    Purple
```
petro-ushchuk, line 9: Enum constants should be upper case and snake case styled.

petro-ushchuk (changes requested):

Andriy-dz (approved):

PR #724, opened 2023-01-28
solution jv-lottery

potrap:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
+    YELLOW, GREEN, RED, BLUE, BLACK, WHITE
```
karma-o, line 4: maybe not a mistake, but in my opinion, it is better to put each on the new line

karma-o (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final int bound = 100;
```
karma-o, line 8: this is a constant, add `static` and name it accordingly 
https://checkstyle.sourceforge.io/reports/google-java-style-20170228.html#s5.2.4-constant-names
```suggestion
    private final int bound = 100;
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color.toString().toLowerCase() + '\''
```
sarakhmen, line 31: It's better to use name() method for getting String representation of enum constants as described [here](for getting String representation of enum constants)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
sarakhmen, line 6: It's better to not create several instances of the same class if you can use only one instance for your purposes :)

sarakhmen (approved):

PR #727, opened 2023-02-02
implemented a simple lottery class, that creates random balls

denys-mg:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
 public class Application {
+    private static Lottery lottery = new Lottery();
```
KseniiaMakarova, line 4: let's better declare it in method main

KseniiaMakarova (approved):
good job!

PR #733, opened 2023-02-05
Created Color enum

VitaliiPotseluiko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
         // create three balls using class Lottery and print information about them in console
```
karma-o, line 9: we can remove this comment now since we have already done what was described here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].toString();
```
karma-o, line 7: inlining operations is not always useful
Here you are creating a new `Random` object every time you are calling this method, let's avoid that by putting this `Random` to a class-level variable
Also worth to mention that it is recommended to call `name()` method for getting the `String` format of that enum instead of `toString()` try to think or google why

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(101));
```
karma-o, line 8: same thing as with Random, but with `ColorSupplier` and `Random` this time

karma-o (changes requested):

src/main/java/core/basesyntax/Application.java
VitaliiPotseluiko: fixed. Please check

src/main/java/core/basesyntax/ColorSupplier.java
VitaliiPotseluiko: fixed. Please check!

src/main/java/core/basesyntax/Lottery.java
VitaliiPotseluiko: fixed. Please check!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(101));
```
sarakhmen, line 10: Please, avoid using magic numbers next time

sarakhmen (approved):

src/main/java/core/basesyntax/Lottery.java
VitaliiPotseluiko: Ok. I will try)

PR #735, opened 2023-02-06
Homework

Skrudj:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        Random random = new Random();
```
sarakhmen, line 7: It's better to use SecureRandom instead and make it a class-level variable

sarakhmen, line 7: It's better to use SecureRandom instead and make it a class-level variable

sarakhmen, line 7: It's better to use SecureRandom instead and make it an instance-level variable.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
```
sarakhmen, line 8: The same problem as in the previous comment. ColorSupplier has an unintuitive name and also should be an instance-level variable.

sarakhmen, line 8: The same problem as in the previous comment. ColorSupplier has an unintuitive name and also should be an instance-level variable.

sarakhmen, line 8: The same problem as in the previous comment. ColorSupplier has an unintuitive name and also should be an instance-level variable.

sarakhmen, line 8: The same problem as in the previous comment. ColorSupplier has an unintuitive name and also should be an instance-level variable.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
+        return new Ball(color.getRandomColor(), random.nextInt(100));
```
sarakhmen, line 9: Please avoid using magic numbers

sarakhmen (changes requested):
Good solution! Let’s improve it 😉

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.security.SecureRandom;
+
+public class Lottery {
+    private static final int RANDOMIZERBOUND = 100;
```
maxline, line 6: Please try to follow the naming convention for constants. Please see examples for constants names that consist from several words

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Lottery lot = new Lottery();
```
maxline, line 5: Please be more accurate with naming variables. We shouldn't  cut and abbreviate the words instead of using whole words. Take a look at examples in the course materials.

maxline (changes requested):
Good job, it is almost complete! Just a couple of minor remarks.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
...
+public enum Colors {
+    Blue,
+    Red,
+    Yellow,
+    Black,
+    White,
+    Gray,
+    Orange,
+    Brown,
+    Green
```
karma-o, line 12: ENUM values should be in uppercase

karma-o (approved):
Approved leaving some naming comment assuming that you will fix it :wink:

PR #736, opened 2023-02-06
Class Ball, Lottery and methods was added

GalubaGSV:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
```
Andriy-dz, line 6: magic number. I think we don't need this variable. you can write logic right away in sout without creating a variable

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    @Override
+    public String toString() {
+        return "Boll color: " + color + ", boll number:" + number;
+    }
```
Andriy-dz, line 10: move this method down

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        Random randomCollor = new Random();
```
Andriy-dz, line 7: just name this variable random and make it a constant

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
```
Andriy-dz, line 6: this method should return String, not a Color

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
...
 
 public class Application {
+    private static final int FIGURE_COUNT = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[FIGURE_COUNT];
+        for (int i = 0; i < FIGURE_COUNT; i++) {
+            balls[i] = new Lottery().getRandomBall();
+            System.out.println(balls[i].toString());
```
KseniiaMakarova, line 10:
```suggestion
            System.out.println(balls[i]);
```
toString() is called automatically

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int FIGURE_COUNT = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[FIGURE_COUNT];
+        for (int i = 0; i < FIGURE_COUNT; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
KseniiaMakarova, line 9: it is better to create Lottery object once before the loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random RANDOM = new Random();
+    private static final int MAX_NUMBER = 100;
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
```
KseniiaMakarova, line 8:
```suggestion
    private static final int MAX_NUMBER = 100;
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

KseniiaMakarova (changes requested):
good job!

KseniiaMakarova (approved):

PR #737, opened 2023-02-06
Hw

copypasterr:
A simple lottery that will return balls with a random color and random number

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,13 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        int ballsQuantity = 3;
+        Ball[] balls = new Ball[ballsQuantity];
+        for (int i = 0; i < ballsQuantity; i++) {
+            balls[i] = new Lottery().getRandomBall();
+        }
+        for (Ball bal: balls) {
+            System.out.println(bal.toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,13 +3,13 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
-        int ballsQuantity = 3;
-        Ball[] balls = new Ball[ballsQuantity];
+        final int ballsQuantity = 3;
+        Ball[] lotteryBalls = new Ball[ballsQuantity];
         for (int i = 0; i < ballsQuantity; i++) {
-            balls[i] = new Lottery().getRandomBall();
+            lotteryBalls[i] = new Lottery().getRandomBall();
         }
-        for (Ball bal: balls) {
-            System.out.println(bal.toString());
+        for (Ball ball: lotteryBalls) {
+            System.out.println(ball.toString());
         }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,33 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball(Color color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+    public Ball () {
+    }
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    public Color getColor() {
+        return color;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball color : " + color.name() + "\nNumber: " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -4,12 +4,14 @@ public class Ball {
     private Color color;
     private int number;
 
+    public Ball() {
+    }
+
     public Ball(Color color, int number) {
         this.color = color;
         this.number = number;
     }
-    public Ball () {
-    }
+
     public void setColor(Color color) {
         this.color = color;
     }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -4,7 +4,7 @@ import java.util.Random;
 
 public class ColorSupplier {
     public Color getRandomColor() {
-        int random = new Random().nextInt(Color.values().length);
-        return Color.values()[random];
+        int colorNumber = new Random().nextInt(Color.values().length);
+        return Color.values()[colorNumber];
     }
 }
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,13 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int ballsQuantity = 3;
+        Ball[] lotteryBalls = new Ball[ballsQuantity];
+        for (int i = 0; i < ballsQuantity; i++) {
+            lotteryBalls[i] = new Lottery().getRandomBall();
```
maxline, line 9: Do we need a create a new instance of `Lottery` every time in a loop?

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,13 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int ballsQuantity = 3;
+        Ball[] lotteryBalls = new Ball[ballsQuantity];
```
maxline, line 7: Do we really need to store the balls in the array or we can just print info in the loop when we creating the ball?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball() {
```
maxline, line 7: do we call the default constructor anywhere in the code?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+        this.number = number;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball color : " + color.name() + "\nNumber: " + number;
```
maxline, line 33: probably would be better to have one line string here. please try to find examples of `toString()` method in course materials or just google its.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int colorNumber = new Random().nextInt(Color.values().length);
```
maxline, line 7: Do we really need to create a new instance of `Random`  every time when this method is called?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(maxNumber));
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
```
maxline, line 6: Please check naming convention for constants in the course materials.
Also try think more accurate naming that express the meaning of the value that constant defines.

SerhiiMykhliuk (changes requested):
Good start! Let's try to improve the result a bit. Please take a look at comments above.

src/main/java/core/basesyntax/Ball.java
copypasterr: It seems at the beginning

src/main/java/core/basesyntax/Application.java
copypasterr: Fixed

src/main/java/core/basesyntax/Lottery.java
copypasterr: Fixed

src/main/java/core/basesyntax/ColorSupplier.java
copypasterr: Fixed

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -3,8 +3,9 @@
 import java.util.Random;
 
 public class ColorSupplier {
+    private final Random randomNumber = new Random();
```
SerhiiMykhliuk, line 6: do we need a `number`  in name of this variable?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,9 +3,11 @@
 import java.util.Random;
 
 public class Lottery {
-    private static final int maxNumber = 100;
+    private static final int RANDOM_RANGE_LIMIT = 100;
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
```
SerhiiMykhliuk, line 7: It is better if `ColorSupplier`  be non-static variable, it is just usual property. The same for the `Random` property. Do not forget to change the uppercase letters when renaming.

src/main/java/core/basesyntax/Ball.java
SerhiiMykhliuk: probably you meant that it was declared in the beginning of the class. 
When I say `called` I mean the place somewhere in the code when we use this constructor  and create a new instances.

src/main/java/core/basesyntax/Lottery.java
maxline, line 7: It is better if `ColorSupplier` be non-static variable, it is just usual property. The same for the `Random` property. Do not forget to change the uppercase letters when renaming.

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        int ballQuantity = 3;
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < ballQuantity; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
maxline, line 9: `toString()` is redundant, it will be called automatically under the hood

copypasterr: deleted

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
KseniiaMakarova, line 5: let's remove a comment

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
...
+        int ballQuantity = 3;
```
KseniiaMakarova, line 6: let's better make this a constant field

KseniiaMakarova (changes requested):
good job! added minor comments

karma-o (approved):

PR #740, opened 2023-02-07
created classes Lottery, Ball, Color; impl methods for random choice of ball's color and number

EliSoll:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
maxline, line 6: Do we need a create a new instance of Lottery every time in a loop?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Color[] colors = Color.values();
+
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(colors.length);
```
maxline, line 9: Do we really need to create a new instance of `Random` every time when this method is called?
It is better to create a new `Random` field as class property.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_VALUE = 101;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(MAX_BALL_VALUE));
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Color[] colors = Color.values();
```
maxline, line 6: Do we really need this property?

PR #741, opened 2023-02-07
jcoop lottery solution

imhatep:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private Colors color;
+
+    public Ball(Colors color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball [color=" + color + ", number=" + number + "]";
```
Andriy-dz, line 14: there should be curly braces {} and without space

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private int index = new Random().nextInt(Colors.values().length);
```
Andriy-dz, line 6: here the variable 'index' is initialized only once and accordingly the method always returns the same value. Let's make only the Random constant, write the name in capital letters and the logic in the middle of the method

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+
+    public Colors getRandomColor() {
```
Andriy-dz, line 8: method should return a String, not a Color

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private int random = new Random().nextInt(100);
```
Andriy-dz, line 6: here the variable 'randon' is initialized only once and accordingly the method always returns the same value. Let's make only the Random constantб write the name in capital letters and the logic in the middle of the method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Ball ball = new Ball(colorSupplier.getRandomColor(),random);
```
Andriy-dz, line 8: lets move this logic in the middle of the method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private int random = new Random().nextInt(100);
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
Andriy-dz, line 7: let's make this variable a constant and write the name in capital letters

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int LOOP = 3;
```
KseniiaMakarova, line 4:
```suggestion
    private static final int BALL_COUNT = 3;
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random RANDOM = new Random();
+    private int index = RANDOM.nextInt(Colors.values().length);
```
KseniiaMakarova, line 7: the problem is that when you create `new ColorSupplier()` - this variable is initialized and is no further changed
so your number is not randomly generated each time you call the `getRandomColor()` method on the same ColorSupplier object - and this is a mistake

please move this variable to the method - so it could be reset each time you call the method

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum name should better be singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random RANDOM = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        int random = RANDOM.nextInt(100);
```
KseniiaMakarova, line 10: let's move 100 to a constant field UPPER_BOUND - because it is a magic number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        return new Ball(new ColorSupplier().getRandomColor(),random);
```
KseniiaMakarova, line 11: why new ColorSupplier()? let's use your field
```suggestion
        return new Ball(colorSupplier.getRandomColor(), random);
```

KseniiaMakarova (changes requested):
please next time be attentive to the branch where you work
you should NOT work in master branch and make a PR from master

KseniiaMakarova (approved):

PR #745, opened 2023-02-07
Application.main will now generate three random balls

De1eF:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int numberOfBalls = 3;
```
KseniiaMakarova, line 4: which naming pattern should constants have?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(100));
```
KseniiaMakarova, line 8: 100 is a magic number - let's make it a constant field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
KseniiaMakarova, line 6: let's not make this method static
instead create a Lottery object in `main()`

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public enum Color {
+    red,
+    yellow,
+    green,
+    blue,
+    purple;
+
+    public Color getValueByInt(int index) {
```
KseniiaMakarova, line 10: let's remove this method

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public enum Color {
+    red,
+    yellow,
+    green,
+    blue,
+    purple;
```
KseniiaMakarova, line 8: enum values should be named as constants

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Color color = Color.red;
+        return color.getValueByInt(new Random().nextInt(5)).toString();
```
KseniiaMakarova, line 8: enum has static method `values()` which returns an array of all its values
let's remove 5 and use this array's length instead, then use random number as index to get the random color from the array

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+            case 0:
+                text = "You got a new ball with ";
+                break;
+            case 1:
+                text = "A strange sphere appeared before you with ";
+                break;
+            case 2:
+                text = "You have won a ball with ";
+                break;
+        }
```
KseniiaMakarova, line 28: overall toString() should be consistent and return information about objects that is consistently formatted
let's remove randomizing from here?

KseniiaMakarova (changes requested):

PR #750, opened 2023-02-07
Lottery logic implementation

mishok321:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
```
sarakhmen, line 7: Please, don't create several instances of the same class if you can use only one instance for your purposes, as described [here](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(Colors.values().length);
```
sarakhmen, line 7: A new Random class instance is created each time the method is called. It's bad practice to initialize dependencies in the method. You can initialize it in fields, but good practice initialization using a constructor.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
sarakhmen, line 3: Enums in Java should be singular.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANDOM_LIMIT = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(RANDOM_LIMIT));
```
sarakhmen, line 9: The same issue with dependencies initialization.

sarakhmen (changes requested):
Let's improve your solution ;)

sarakhmen (approved):
LGTM! 🚀

PR #753, opened 2023-02-08
Main

OlegMR0:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
```
kozhukhovsky, line 4: You don't need three lottery objects. One is enough. Also, since it is one, you can call it 3 times in a loop. 
And you don't need to call toString() explicitly, because it is called implicitly at objects in println() method

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public Ball(String color,int num) {
+        this.color = color;
+        this.num = num;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + getColor() + " \n" + "Number: " + getNum() + "\n";
+    }
```
kozhukhovsky, line 15: move toString() method after getters

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].toString();
```
kozhukhovsky, line 7: It is better to put random into a class variable, as in the example: 

image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        return new Ball(color.getRandomColor(), new Random().nextInt(100));
```
kozhukhovsky, line 8: The same. Put colorSupplier and random like a class variable.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
```
kozhukhovsky, line 7:
```suggestion
        ColorSupplier colorSupplier = new ColorSupplier();

```

src/main/java/core/basesyntax/Lottery.java
kozhukhovsky, line 8: 100 is magic number. Move him to the constant.

image

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Ball.java
OlegMR0: Okay, but why? Is it just a convention? It's just for myself.

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
...
         // create three balls using class Lottery and print information about them in console
+        Lottery ball = new Lottery();
```
nacenik, line 6: It isn't a ball. it's lottery
```suggestion
        Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int number = new Random().nextInt(Color.values().length);
```
nacenik, line 7: Random  should be class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
+
+    public Ball getRandomBall() {
+
```
nacenik, line 9: redundant empty line

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
+
+    public Ball getRandomBall() {
+
+        ColorSupplier colorSupplier = new ColorSupplier();
```
nacenik, line 10: ColorSupplier should be class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
+
+    public Ball getRandomBall() {
+
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String colorStr = colorSupplier.getRandomColor();
+        int number = new Random().nextInt(BOUND);
```
nacenik, line 12: Random should be class level variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Color.values()[number].toString();
```
nacenik, line 8: [use `name()` to get string representation of enum values](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
```
nacenik, line 6: add access modifier

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
...
+        return color;
+    }
+
+    public int getNum() {
+        return num;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + getColor() + " \n" + "Number: " + getNum() + "\n";
```
nacenik, line 22: you have direct access to variables inside the class. 
```suggestion
        return "Color: " + color + " \n" + "Number: " + number + "\n";

```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int num;
```
nacenik, line 5: Use the short form of words for class variables naming - bad practice
```suggestion
    private int number;
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
...
+        this.num = num;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public int getNum() {
+        return num;
+    }
```
nacenik, line 18:
```suggestion
    public int getNumber() {
        return number;
    }

```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int num;
+
+    public Ball(String color,int num) {
+        this.color = color;
+        this.num = num;
```
nacenik, line 9:
```suggestion
    public Ball(String color,int number) {
        this.color = color;
        this.number = number;

```

nacenik (changes requested):

src/main/java/core/basesyntax/Lottery.java
OlegMR0: Are you sure there?

I've tested and it outputted me every time same object.

OlegMR0: Same with number

nacenik: I'm 100% sure about it

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private int number = new Random().nextInt(Color.values().length);
+
```
nacenik, line 7:
```suggestion
    private Random random = new Random()
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
     public String getRandomColor() {
-        return null;
+        return Color.values()[number].name();
```
nacenik, line 9:
```suggestion
        private int number = random.nextInt(Color.values().length);
        return Color.values()[number].name();
```

src/main/java/core/basesyntax/Lottery.java
nacenik: you need to change ColorSupplier as I proposal

src/main/java/core/basesyntax/Lottery.java
nacenik: not fixed. Do it like in ColorSupplier

nacenik (changes requested):

src/main/java/core/basesyntax/Lottery.java
okuzan: not fixed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int BOUND = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private int number = new Random().nextInt(BOUND);
```
okuzan, line 8: this should not be a field, only BOUND should. This defeats the purpose of the class - Lottery won't return different balls, this value only will be set once

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: better delete comments too

okuzan (changes requested):

aaananas (changes requested):
image

PR #755, opened 2023-02-08
added Ball Color Lottery classe and implemented all necessary methods

OlehSarapuk:
what is a better practice in Lottery class 9 - 13 lines

my implenetation:
`ColorSupplier colorSupplier = new ColorSupplier();
 Random random = new Random();
 String randomColor = colorSupplier.getRandomColor();
 int randomNumber = random.nextInt(FIGURE_COUNT);
 return new Ball(randomColor, randomNumber);`
 
 or something like this:
 `return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(FIGURE_COUNT));`
?

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,32 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "color -> " + color + " number -> " + Integer.toString(number);
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(Color.values().length);
+        return Color.values()[randomIndex].name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int FIGURE_COUNT = 100;
+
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "color -> " + color + " number -> " + Integer.toString(number);
```
sarakhmen, line 30: I'd recommend using IntelliJ IDEA tools to generate toString() method in order to make it more unified and comprehensive :)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(Color.values().length);
```
sarakhmen, line 7: It's bad practice to initialize dependencies in the method. You can initialize it in fields, but good practice initialization using a constructor. So, please, move Random instance initialization from this method.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int FIGURE_COUNT = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
```
sarakhmen, line 10: The same issue, as previously. Please, move ColorSupplier and Random initialization from this method.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+        String randomColor = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(FIGURE_COUNT);
+        return new Ball(randomColor, randomNumber);
```
sarakhmen, line 13: And as for your question, I'd prefer to make use of something like next:
```suggestion
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(FIGURE_COUNT));
```

sarakhmen (changes requested):

OlehSarapuk:
thanks for your answer

i can't understand how to initialize dependencies using a constructor, but i tried to do it somehow
can you please explain me how to do it right)

src/main/java/core/basesyntax/Lottery.java
maxline: Please take a look to the previous comment with suggested solution
` return new Ball(colorSupplier.getRandomColor(), random.nextInt(FIGURE_COUNT));`

maxline (changes requested):
Good job, just one remark is left.

maxline (approved):
Well done! Approved.

PR #759, opened 2023-02-08
HomeWork Lottery/JavaCORE OOP

NickolayVoloshchuk:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
 public class Application {
+    public static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i <= NUMBER_OF_BALLS; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
```
maxline, line 8: toString() is redundant inside println, automatically called under the hood

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndexColor = new Random().nextInt(Colors.values().length);
```
maxline, line 7: should we create new instance of Random every time when the method is called?
create new var `Random` type  as class property and initialize there.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_BALL_NUMBER_RANDOM = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new Random().nextInt(MAX_BALL_NUMBER_RANDOM),
+                        new ColorSupplier().getRandomColor());
```
maxline, line 10: the same as previous comment.
should we create new instance of Random and ColorSupplier  every time when the method is called?
create a class attributes for  Random and ColorSupplier and initialize them at class level, not inside the method block.

maxline (changes requested):
Good job! Let's make it better, please take a look at several comments above.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,10 +3,11 @@
 import java.util.Random;
 
 public class Lottery {
-    public static final int MAX_BALL_NUMBER_RANDOM = 100;
+    private static final int MAX_BALL_NUMBER_RANDOM = 100;
+    private Random randomNumberBall = new Random();
+    private ColorSupplier randomColor = new ColorSupplier();
```
maxline, line 8: it is better to name like 'colorSupplier'

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,10 +3,11 @@
 import java.util.Random;
 
 public class Lottery {
-    public static final int MAX_BALL_NUMBER_RANDOM = 100;
+    private static final int MAX_BALL_NUMBER_RANDOM = 100;
+    private Random randomNumberBall = new Random();
```
maxline, line 7: Question about naming. This is just `Random` class instance, probably clarifying it as `NumberBall` is redundant.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -3,8 +3,9 @@
 import java.util.Random;
 
 public class ColorSupplier {
+    private Random randomIndexColor = new Random();
+
     public String getRandomColor() {
-        int randomIndexColor = new Random().nextInt(Colors.values().length);
-        return Colors.values()[randomIndexColor].name();
+        return Colors.values()[randomIndexColor.nextInt(Colors.values().length)].name();
```
maxline, line 9: Please use singular names for  enums and classes. `Color` instead of `Colors`

maxline (changes requested):
Well done! Fix please last small remarks.

maxline (approved):
Well done! Approved

PR #763, opened 2023-02-08
Added classes 'Ball', 'Colors', 'Lottery' and implemented methods

AndrewkeY-hub:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
petro-ushchuk, line 6: Magic number `3` should be constant

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].toString();
```
petro-ushchuk, line 7: You don't need to create new `Random` instance every time you call `getRandomColor`, introduce it as field variable `random`.
And then provide `random.nextInt(Colors.values().length)` as local variable to make your code more readable.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
petro-ushchuk, line 3: Enum naming should be singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANDOM_COUNT = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(RANDOM_COUNT));
```
petro-ushchuk, line 10: You don't need to create new `Random` instance every time you call `getRandomBall`, introduce it as field variable random.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int THREE = 3;
+    private static final int ZERO = 0;
```
karma-o, line 5: constants should have a context meaning of that variable, and answer the questions like what is this number for? Or what does this number represent?
```suggestion
    private static final int MAX_BALLS = 3;
    private static final int START_INDEX = 0;
```

karma-o (changes requested):

karma-o (approved):

karma-o (changes requested):
Sorry but you have a build failure because of the checkstyle, we need to fix it

AndrewkeY-hub:
There were some extra spaces, already fixed

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int MAX_BALLS = 3;
+    private static final int START_INDEX = 0;
```
sarakhmen, line 5: It's pretty expected that array is starting from 0 index, so you please don't worry about it

sarakhmen (approved):

PR #764, opened 2023-02-08
Java lottery home work

V-Vasylchuk:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Ball {
+    public String color;
+    public int number;
```
maxline, line 5: Please use private modifiers for class fields and create getters for them

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
maxline, line 7: We don't need create a new class instance for `Random` every time when method `getRamdomColor()` is called. Move it to the class field please and initialize just one time.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[index].name();
```
maxline, line 8: Please use singular names for enums and classes. `Color` instead of `Colors`.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER_VALUE = 100;
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(MAX_NUMBER_VALUE));
```
maxline, line 8: We don't need create a new instances for classes `Random` and `ColorSupplier` every time when method `getRamdomBall()` is called. Move them to the class fields please and initialize just one time.

maxline (changes requested):
Good job, mate! Let's make a solution even better. Please take a look at comments above.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
KseniiaMakarova, line 6: it is better to create a lottery object once before the loop

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    Random random = new Random();
```
KseniiaMakarova, line 6: what about access modifier?
add an empty line between fields and methods

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER_VALUE = 100;
+    Random random = new Random();
+    ColorSupplier colorSupplier = new ColorSupplier();
```
KseniiaMakarova, line 8: same: access modifiers + empty line

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):
github check fails: run mvn clean package locally and fix the mistake

sarakhmen (approved):
Good solution :)

PR #766, opened 2023-02-08
Home work lottery

PlagueEbola:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
petro-ushchuk, line 3: Enum name should be singular

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i <= 3; i++) {
```
petro-ushchuk, line 6: Magic number `3` should be constant

petro-ushchuk, line 6: Start your iterator from `0` to `< 3` instead. Remember about `0` index, it can create a lot of problems with read/write operations with arrays.

petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        final int quantityBalls = 3;
```
KseniiaMakarova, line 6: let's make it a constant field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int bound = 100;
```
KseniiaMakarova, line 6: constants should have a PROPER_NAMING pattern

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #767, opened 2023-02-08
Created Color enum and Ball, ColorSuplier and Lottery classes

AnonimToster:


petro-ushchuk (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_COUNT = 100;
+    private Random random;
+    private ColorSupplier colorSupplier;
+
+    public Ball getRandomBall() {
+        Color color = Color.valueOf(colorSupplier.getRandomColor());
```
karma-o, line 11: Lets bring our `Color` representation to a single type, both `ColorSupplier.getRandomColor()` return type and `Ball.color` field type should be the same

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_COUNT = 100;
+    private Random random;
+    private ColorSupplier colorSupplier;
```
karma-o, line 8: Initialize these variables, same with `Random` in `ColorSupplier`

karma-o (changes requested):

karma-o (approved):
:fire:

PR #770, opened 2023-02-08
added lottery solution

en4ik99:


karma-o (approved):
Good job!

karma-o (changes requested):
Sorry about the inconvenience, but you have a build failure after the approval, so we will need to fix that, do not forget to run `mvn clean package`

PR #771, opened 2023-02-08
Maksym Yashyn HomeWork lottery

Nimakel:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery ball = new Lottery();
```
KseniiaMakarova, line 5:
```suggestion
        Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum names should better be singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
KseniiaMakarova, line 9: let's better set fields through a constructor?

KseniiaMakarova, line 9: 100 is a magic number - we should make it a constant field

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        final int number = 100;
```
KseniiaMakarova, line 7: this is not a constant field because you declare it in a method
please move it to the class level, and remember that constants should be static, final, and have PROPER_NAMING pattern

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        Ball ball = new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(number));
```
KseniiaMakarova, line 8: ideally, you should move creation of new objects (new ColorSupplier(), new Random()) outside the method as well (create private fields)

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Lottery.java
Nimakel: Yeah, maybe, ideally i should move creation outside the method, but when i do it, i'm always have the same result:
image
If i'd like to have a random result, like a task want, i still have to create object in method, but now in main() method. I would like to stay it creation of object in this method, cos it's doesn't metter where i will create it.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER = 100;
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(NUMBER));
+        return ball;
+    }
+}
```
KseniiaMakarova, line 12:
```suggestion
public class Lottery {
    private static final int NUMBER = 100;
    private Random random = new Random();
    private ColorSupplier colorSupplier = new ColorSupplier();

    public Ball getRandomBall() {
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(NUMBER));
    }
}
```

that way you shouldn't get same generated values each time, and also you will create only 1 object per program (not each time you enter the method)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int random = new Random().nextInt(Color.values().length);
```
KseniiaMakarova, line 7: same here, try it

KseniiaMakarova (approved):

PR #773, opened 2023-02-08
Implemented Lottery class and helper classes.

Deschna:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
+import java.util.Arrays;
+
 public class Application {
+    public static final int BALLS_COUNT = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[BALLS_COUNT];
+        for (int i = 0; i < balls.length; i++) {
```
karma-o, line 11: since we already have a constant I'd use it instead
```suggestion
        for (int i = 0; i < BALLS_COUNT; i++) {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
karma-o, line 7: `Random` should be reused, so let's move it to a class-level var

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAXIMUM_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(MAXIMUM_NUMBER));
```
karma-o, line 9: same thing but also with a `ColorSupplier` now

karma-o (changes requested):

sarakhmen (approved):
Good solution :)

PR #775, opened 2023-02-08
created three balls with random color and number

pechivo:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
karma-o, line 3: enum class naming should be singular
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
+
+    public Ball getRandomBall() {
+        Ball newBall = new Ball();
+        newBall.setColor(Colors.values()[new Random().nextInt(Colors.values().length)].name());
```
karma-o, line 10: you have a `ColorSupplier`, use it here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
```
karma-o, line 6: I will not count it as a mistake, because operating the actual `Color`  values instead of the `String` representation of it is actually better, but it is also recommended to not change the signature of the method, so this is up to you
And if you have a Color instance here, you might as well have a `Ball.color` field  be of type `Color`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
...
+        return Colors.values()[new Random().nextInt(Colors.values().length)];
```
karma-o, line 7: you are creating a new `Random` instance every time, make it a class-level var

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        newBall.setNumber(new Random().nextInt(BOUND));
```
karma-o, line 11: same thing with `Random`

karma-o (changes requested):

pechivo:
Thanks for review)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
```
sarakhmen, line 9: You have random instance initialized as a field, so why don't use it?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        Ball newBall = new Ball();
+        newBall.setColor(new ColorSupplier().getRandomColor());
```
sarakhmen, line 11: Please, initialize ColorSupplier the same way you did with Random

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
sarakhmen, line 9: Please, don’t create several instances of the Lottery class if you can use only one instance for your purposes. Also, I wouldn't recommend creating a separate array for balls, you can just get a ball using getRandomBall() method and print it straight away.

petro-ushchuk (approved):

PR #783, opened 2023-02-09
add solution

tyryshkin1997:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[3];
+        for (int i = 0; i < 3; i++) {
+            lottery[i] = new Lottery();
+            System.out.println(lottery[i].getRandomBall());
+        }
```
KseniiaMakarova, line 9: why do you have 3 lotteries? there should be 1 lottery and 3 balls

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private int number;
+    private String color;
+
+    public Ball(int number, Colors color) {
+        this.number = number;
+        this.color = String.valueOf(color);
+    }
```
KseniiaMakarova, line 10:
```suggestion
    public Ball(int number, String color) {
        this.number = number;
        this.color = color;
    }
```
please do casting from enum to String in a different place
constructor should not contain logic

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum names should better be singular

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int color = new Random().nextInt(Colors.values().length);
```
KseniiaMakarova, line 7: probably it is better to move new Random() initialization out of this method - like in your Lottery class

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #786, opened 2023-02-09
Application print information about three random balls

s-hrynus:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
Andriy-dz, line 7: it is better to create the Random object as a class field with the final modifier

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Andriy-dz, line 6: magic number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball(colorSupplier.getRandomColor(), new Random().nextInt(100));
```
Andriy-dz, line 8: magic number
it is better to create the Random and ColorSupplier objects as a class field with the final modifier

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Application.java
s-hrynus: as result I must do without loop?))

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
-    public String getRandomColor() {
-        return null;
+    private static final int COLOR_COUNT = Colors.values().length;
+
+    private final Random random = new Random();
+
+    public Colors getRandomColor() {
+        int index = random.nextInt(COLOR_COUNT);
+        return Colors.values()[index];
     }
```
petro-ushchuk, line 13: `getRandomColor` should return `String`, not `Colors`

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
petro-ushchuk, line 3: Enum name should be singular

petro-ushchuk (changes requested):

sarakhmen (approved):
Good solution ;)

PR #793, opened 2023-02-13
1.Created a Ball class with the following fields: color and number an…

TemkaHillfinger:
…d toString() method

2.Create an enum with different colors
implemented the getRandomColor() method in the ColorSupplier class. This method returns a random color based on enum values 3.Created a Lottery class with a getRandomBall() method that returns a ball. This method creates a ball of random color and random number

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return String.valueOf(Color.values()[index]);
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
```suggestion
public class ColorSupplier {
    private final Random random = new Random();
    
    public String getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return String.valueOf(Color.values()[index]);
```
```suggestion
        int index = random.nextInt(Color.values().length);
        return Color.values()[index].name();
```

aaananas (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private final int maxValues = 100;
```
```suggestion
    private static final int MAX_VALUE = 100;
```

aaananas (approved):

src/main/java/core/basesyntax/Lottery.java
TemkaHillfinger: I tried to make static but maven test failed

PR #796, opened 2023-02-13
hw-2-lottery

VolodymyrDavidov:


src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    Red,
```
uliana-andrushchuk, line 4: Enum members are constants. You should use uppercase here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
uliana-andrushchuk, line 8: It's better to use `name()` for getting String representation of enum constants

uliana-andrushchuk, line 8: `Color.values()` can be extracted as separate variable, as it's used twice

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Color[] getColor = Color.values();
+        int index = new Random().nextInt(getColor.length);
+        return getColor[index].name();
```
aleksandr-bevza, line 9: 1. getColor is a good name for method, but not for variable.
2. Move Random to class field. Don't create new Random each time this method is invoked.
```suggestion
        Color[] colorValues = Color.values();
        int index = random.nextInt(colorValues.length);
        return getColor[index].name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER = 100;
```
aleksandr-bevza, line 6: Variable's name should explain it's purpose
```suggestion
    private static final int NUMBER_LIMIT = 100;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
```
aleksandr-bevza, line 9:
```suggestion
```

aleksandr-bevza (changes requested):

PR #797, opened 2023-02-14
implement main method with creating 3 random ball

Rolik-Oleksandr:
Create ENUM class with colors
implement random class in ColorSupplier
implement main method with creating 3 random ball

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
+    private final int number;
```
Ivan95kos, line 5:
```suggestion
    private String color;
    private int number;
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Ivan95kos, line 6:
```suggestion
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
```

PR #800, opened 2023-02-18
jv lottery solution

furthernull:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
...
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = {
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()
+        };
+
+        for (Ball ball: balls) {
+            System.out.println(ball.toString());
```
okuzan, line 13: FYI toString() is called by default

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = {
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()
+        };
+
+        for (Ball ball: balls) {
```
```suggestion
        for (Ball ball : balls) {
```

okuzan: use Intellij shortcut to auto-format before committing - like Ctrl Alt L if you have Windows

src/main/java/core/basesyntax/Application.java
okuzan, line 13: Don't needlessly recreate lottery instances. Create one lottery instance before the for loop, then use that instance in the loop. We don't need to save the ball as a variable at all, just print the result of the .getRandomBall() method invocation

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].toString();
```
okuzan, line 7: make random class field. This way we will create in only once and reuse it

okuzan: here .toString() is a checklist issues, revisit it please

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public enum Colors {
```
okuzan, line 3: Don't call Enum class in plural, it doesn't comply with the convention

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball randomBall = new Ball();
+        randomBall.setColor(new ColorSupplier().getRandomColor());
+        randomBall.setNumber(new Random().nextInt(101));
```
okuzan, line 9: make ColorSupplier, Random class fields, 101 - class constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
```
okuzan, line 10: delete empty line

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = {
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()
+        };
+
```
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: delete comment

okuzan (changes requested):

aleksandr-bevza (approved):

PR #809, opened 2023-02-25
Hw jv lottery

vrudkovsky:


pom.xml
```diff
@@ -64,5 +64,4 @@
             </plugins>
         </pluginManagement>
     </build>
-</project>
-
+</project>
```
Ivan95kos, line 67: No need to edit pom.xml file

Ivan95kos (approved):
Great job

PR #816, opened 2023-02-28
create Ball class with the following fields: color and number and ove…

vromanyuk:
…rride toString() method

create enum with different colors
implement method getRandomColor() in class ColorSupplier. This method should return a random color, based on the enum values (use Random class for this) Then create a Lottery class with getRandomBall() method, which will return the ball. This method should create a ball with a random color and a random number (the maximum possible number should be 100) and return this ball from the method.

To test the program, in main() method, create three balls using getRandomBall() method from Lottery class. And print information about them in the console.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,8 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        int i = new Random().nextInt(colors.length);
+        return colors[i].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
+    RED,
+    GREEN,
+    BLUE,
+    YELLOW,
+    ORANGE
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall(){
+        int index = new Random().nextInt(100);
+        return new Ball(new ColorSupplier().getRandomColor(), index);
+    }
+}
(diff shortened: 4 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
```
kozhukhovsky, line 7: you don't need create new Lottery each iteration, move this to the variable before loop and use it in loop

kozhukhovsky: It is not necessary to call toString explicitly, because in the print method toString is called implicitly on objects

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        int i = new Random().nextInt(colors.length);
```
kozhukhovsky, line 8: move random to the class constant, like:
image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall(){
+        int index = new Random().nextInt(100);
```
kozhukhovsky, line 7: move random and color supplier to the class constant

kozhukhovsky, line 7: magic number
image

kozhukhovsky (changes requested):

Ivan95kos (changes requested):
fix all checks
Знімок екрана 2023-02-28 о 20 01 35

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: delete comments before committing

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+    private final Color[] colors = Color.values();
+
     public String getRandomColor() {
-        return null;
+        int i = random.nextInt(colors.length);
+        return colors[i].toString();
```
okuzan, line 11: [make sure you go thru checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
vromanyuk: I don't understand, where is the problem

vromanyuk: You indicated on the problem "Use name() for getting String representation of enum constants"

okuzan (approved):
Well done!

PR #824, opened 2023-03-07
3 random-coloured and random-numbered balls implemented.

TymurProkhorov:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = new Lottery().getRandomBall();
```
okuzan, line 7: you need only one lottery instance, and you don't need to create ball variable explicitly, print the result of the call direclty

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
okuzan, line 7: random should be class field

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
okuzan, line 3: enum name should be in singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
okuzan, line 7: make ColorSupplier, Random class fields, and 100 class constant

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
okuzan, line 7: this will create 3 lottery instances. you need to be efficient - create one and reuse it in the loop

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,21 @@
...
+
+    public Random getRandom() {
+        return random;
+    }
+
+    public void setRandom(Random random) {
+        this.random = random;
     }
 }
+
```
okuzan, line 21: delete this line

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,32 @@
...
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Random getRandom() {
+        return random;
+    }
+
+    public void setRandom(Random random) {
+        this.random = random;
+    }
```
okuzan, line 31: why would random and color supplier be set or retrieved? they are private for a reason. This class is not a case to create getters/setters

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: delete comments before committing

okuzan (changes requested):

aaananas (approved):

PR #828, opened 2023-03-08
Hw oop/lotery

andrii-subash:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int numberLimit = 100;
+
+    private Random random = new Random();
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
olekskov, line 10:
```suggestion
    private static final int NUMBER_LIMIT = 100;
    
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

```

olekskov (approved):
Good job, approved.
Please pay attention to comment below for further tasks. Magic numbers are better as constants and some services on which we depend are better to declare final
And also don't put empty line between each variable. One gap between constants is ok, but not between variables unless really improves readability in rare cases

PR #841, opened 2023-03-19
Complete homework v1.0.0

eugenetolkunov:


src/main/java/core/basesyntax/ColorBall.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum ColorBall {
```
rovein, line 3: Let it be just `Color`.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Random;
+
+public class Lottery {
+    public void getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setColor("" + colorSupplier.getRandomColor());
+        Random random = new Random();
+        ball.setNumber(random.nextInt(100));
+        System.out.println(ball);
```
rovein, line 12: Don't create `ColorSupplier` and `Random` here, instead move it to the class-level.

Also, don't use `"" + colorSupplier.getRandomColor()`, instead call method `name()`. [Reference](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants).

And don't. use magic number `100`.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        lottery.getRandomBall();
+        lottery.getRandomBall();
+        lottery.getRandomBall();
```
rovein, line 8: Use loop to create several objects, see [here](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-for-loop-for-creating-several-objects-of-the-same-class). And move the `System.out.println()` here.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public ColorBall getRandomColor() {
+        int index = new Random().nextInt(ColorBall.values().length);
+        return ColorBall.values()[index];
```
rovein, line 8: Don't create `Random` here, move it to the class-level declaration.

rovein (changes requested):
Good solution! But let's improve it ;)

sarakhmen (approved):
good job :)

PR #846, opened 2023-03-24
Made a solution

anna-poplavska:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: better delete comments before committing

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
...
+        final int numberOfBalls = 3;
```
okuzan, line 6: make it class constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
...
+        Ball[] balls = new Ball[numberOfBalls];
+        for (int i = 0; i < numberOfBalls; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
okuzan, line 9: main can do with only one lottery instance, let's be efficient, what if it was costly operation

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
...
+            System.out.println(balls[i].toString());
```
okuzan, line 10: toString() is called by default is system out println

src/main/java/core/basesyntax/Application.java
okuzan, line 9: we don't need to save balls to array, we can print them right after getting from the method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
okuzan, line 7: Random, ColorSupplier should be class fields with appropriate access modifiers, 100 should be class constant

okuzan (changes requested):
Good job so far!

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    static final int numberOfBalls = 3;
```
okuzan, line 4: should be private and in CONSTANT_CASE

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
...
+    private static Ball ball;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < numberOfBalls; i++) {
+            ball = new Lottery().getRandomBall();
```
okuzan, line 9: create lottery outside of a loop

```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
```
okuzan, line 6: CONSANT_CASE

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(maxNumber));
```
okuzan, line 9: ColorSupplier and Random should be class fields - places after the constant : Random random = new  Random(); etc

okuzan (changes requested):

okuzan (approved):
Well done!

PR #849, opened 2023-03-25
lottery-solution

z1row:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = {lottery.getRandomBall(), lottery.getRandomBall(), lottery.getRandomBall()};
```
olekskov, line 6: It's better to fill array also in cycle for the case tomorrow we will be asked to make 1000 balls

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        int number = new Random().nextInt(100);
```
olekskov, line 9: Please extract Random to class variable as done in ColorSupplier
Please make all magic numbers as class constants with proper naming

olekskov (changes requested):

PR #852, opened 2023-03-27
created class ball, created Enum with colors, created Lottery class a…

OlehPryshchepa:
…nd implement methods

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,9 +3,11 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
-    private int index = new Random().nextInt(Colors.values().length);
+    private int index;
+    private Random random = new Random();
 
     public String getRandomColor() {
+        index = random.nextInt(Colors.values().length);
         return Colors.values()[index].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -3,9 +3,13 @@ package core.basesyntax;
 import java.util.Random;
 
 public class Lottery {
-    private Ball ball = new Ball(new Random().nextInt(100), new ColorSupplier().getRandomColor());
+    private static final int MAX_SIZE = 100;
+    private Ball ball;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
 
     public Ball getRandomBall() {
+        ball = new Ball(random.nextInt(MAX_SIZE), colorSupplier.getRandomColor());
         return ball;
     }
 }


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[]{
+                new Lottery().getRandomBall(),
```
maxline, line 6: Do we need a create a new instance of Lottery every time in a loop?
https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
maxline, line 7: Do we really need to create a new instance of Random every time when this method is called?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new Random().nextInt(100), new ColorSupplier().getRandomColor());
```
maxline, line 7: The same as previous comment.
Do we really need to create a new instance of ColorSupplier and Random classes every time when this method is called?

maxline (changes requested):
Good job! let's improve the solution

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private int index = new Random().nextInt(Colors.values().length);
```
sarakhmen, line 6: your index has the same value on each getRandomColor() call. Let's better calculate index value within the method and make your Random instance a class field (initialized directly in the field or using a constructor)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Ball ball = new Ball(new Random().nextInt(100), new ColorSupplier().getRandomColor());
```
sarakhmen, line 6: The same issue here. Also, '100' is a magic number, please read common mistakes carefully. ColorSupplier is also your dependency and should be a field

rovein (approved):
Good job! All previous comments have fixed.

PR #853, opened 2023-03-27
Lottery solution

z1row:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
```
uliana-andrushchuk, line 12:
```suggestion
        for (int i = 0; i < balls.length; i++) {
            System.out.println(lottery.getRandomBall());
        }
```
Can be simplified to one loop

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
+    private final int num;
```
uliana-andrushchuk, line 5:
```suggestion
    private final int number;
```
Its better to use full names for variables

uliana-andrushchuk (approved):

PR #858, opened 2023-04-02
jv-lottery solution

avereshchynska:


olekskov (changes requested):
build is failed, please run 'mvn clean package' command locally and fix codestyle issues in report to be able to proceed with review

avereshchynska:
i've run mvn clean package and fixed errors

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+
+        System.out.println(ball1.toString());
```
okuzan, line 10: FYI, toString will be called be default, no need to do it explicitly

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: better delete comments

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 7: in your implementation, Random will be created each time the method is called. Make it a class field, this will make the code more efficient

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private int CONSTANT_CASE = 100;
```
uliana-andrushchuk, line 8:
```suggestion
    private static final int CONSTANT_CASE = 100;
```
Please, rename this variable
using CONSTANT_CASE means that all constants should be in upper case and use underscore
but the variable name should give information about what it is used for
(ex `MAX_BALL_NUMBER`)

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/Lottery.java
avereshchynska: oh, sorry. it was a hard day.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
okuzan, line 7: it's a local variable, still created each time the method is called, move it out of the method, don't forget appropriate access modifier

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private int MAX_POSSIBLE = 100;
```
okuzan, line 8:
```suggestion
    private int MAX_NUMBER = 100;
```

okuzan: to make it constant, it also should be static and final

okuzan: your checkstyle doesn't pass because checker thinks it's wrong case for a field

okuzan (changes requested):
Please always make sure your build passes before requesting a review

src/main/java/core/basesyntax/Lottery.java
avereshchynska: run mvn clean package and fix errors.

okuzan (approved):
Nice job!

PR #871, opened 2023-04-15
Hw jv lottery solution

kvjohnny:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public void setColour(String colour) {
```
olekskov, line 7: The method name should be `setColor` instead of `setColour` to match the variable name.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: Consider renaming the enum to 'Color' instead of 'Colors' to follow Java naming conventions for enums. Enums are usually named in singular form. 🤓

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
-public class ColorSupplier {
+import java.util.Random;
+
+public abstract class ColorSupplier {
```
olekskov, line 5: Remove the `abstract` keyword from the class declaration, as there is no need for this class to be abstract. 🧐

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    private static final int MAX_NUMBER_OF_BALLS = 100;
+
+    @Override
```
olekskov, line 8: The getRandomColor() method should be removed from Lottery class and used from the ColorSupplier instance mentioned earlier.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+    public String getRandomColor() {
+        return Colors.values()[new Random().nextInt(Colors.values().length)].toString();
+    }
+
+    public String getRandomBall() {
```
olekskov, line 13: getRandomBall() should return a Ball object instead of a String. You should create a new Ball instance with the random color and random number, and return it. Example: return new Ball(colorSupplier.getRandomColor(), RANDOM.nextInt(MAX_BALL_NUMBER));

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
```
olekskov, line 3: Please create a single instance of Random at the class level, rather than creating a new instance each time it's used. This helps improve performance and code readability. Example: private static final Random RANDOM = new Random();

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class Lottery extends ColorSupplier {
```
olekskov, line 5: Lottery should not extend ColorSupplier, instead create an instance of ColorSupplier and use it in the getRandomBall() method. Example: private final ColorSupplier colorSupplier = new ColorSupplier();

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,41 @@
...
+        Ball secondBall = new Ball();
+        secondBall.setColour("RED");
+        secondBall.setNumber(81);
+        System.out.println(secondBall);
+
+        Ball thirdBall = new Ball();
+        thirdBall.setColour("BLUE");
+        thirdBall.setNumber(4);
+        System.out.println(thirdBall + "\n");
+        System.out.println("---------------------------------\n");
```
olekskov, line 32: This part is not relevant for the task
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,41 @@
...
+
+        System.out.println("Lottery results: ");
+        ColorSupplier[] balls = new Lottery[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery();
+            System.out.println(balls[i].getRandomBall());
+        }
```
olekskov, line 39: After you implement other changes your method should look something like below:
```suggestion
        System.out.println("Lottery results: ");
        Lottery lottery = new Lottery();
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
kvjohnny: According to task: "create a Lottery class with getRandomBall() method". Such actions will be contrary to this task condition.

src/main/java/core/basesyntax/Lottery.java
kvjohnny: I changed field from "private static final Random RANDOM = new Random();" to "private final Random random = new Random();", because after execution command "mvn clean package" I got next error message: "Static variable definition in wrong order".
Static RANDOM

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", number=" + number
+                + '}';
+    }
+}
+
+
```
kozhukhovsky, line 29: remove empty lines

kozhukhovsky (approved):

PR #873, opened 2023-04-18
task solution implemented

Tupadzaky:


src/main/java/core/basesyntax/suppliers/ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax.suppliers;
+
+import core.basesyntax.enums.Color;
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
```
sarakhmen, line 8: Here, your random instance is created every time the method is called. random - is our dependency, let's initialize dependencies using a constructor or directly create a new instance in the class fields.

src/main/java/core/basesyntax/suppliers/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.suppliers;
+
+import core.basesyntax.model.Ball;
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier supplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(supplier.getRandomColor(), new Random().nextInt(100));
```
sarakhmen, line 10: the same for random. Also, let's avoid using magic numbers, as described in the common mistakes

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import core.basesyntax.model.Ball;
+import core.basesyntax.suppliers.BallSupplier;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        BallSupplier ballSupplier = new BallSupplier();
+        Ball[] balls = ballSupplier.createBalls(3);
```
sarakhmen, line 9: Number of balls here is also a magic number :)

sarakhmen (changes requested):
Overall good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/suppliers/BallSupplier.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax.suppliers;
+
+import core.basesyntax.model.Ball;
+
+public class BallSupplier {
```
maxline, line 5: Looks like this class is not only supplier, it can also print balls. So I think it is better rename class as BallService

src/main/java/core/basesyntax/model/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax.model;
+
+public class Ball {
+    private final String color;
```
maxline, line 4: let's use existing `enum Color` instead of `String` for this field

maxline (changes requested):
Good job! let's improve the solution

rovein (approved):
Good job! Keep the rhythm!

PR #874, opened 2023-04-18
task solved

MykytaKat:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        Random random = new Random();
+        int index = random.nextInt(colors.length);
+        return String.valueOf(colors[index]);
```
sarakhmen, line 10: As described in common mistakes, it's better to use name() method in order to get string representation of enum

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
+    Red,
+    Green,
+    Yellow,
+    Blue,
+    Black,
+    White
```
sarakhmen, line 9: yours enums should be uppercased

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int value = random.nextInt(100);
```
sarakhmen, line 8: as described in common mistakes, let's avoid using magic numbers

sarakhmen (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
sarakhmen, line 6: '3' is also magic number here

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        int numberOfBalls = 3;
```
sarakhmen, line 5: should be class constant (private static final)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier color;
+    private Random random;
+    private int bound = 100;
```
sarakhmen, line 8: should be also static and final. And do not forget that class constants should be uppercased

sarakhmen (changes requested):
Let's improve your solution :) Also let's reduce the number of commits if they are sequential and have a small impact

src/main/java/core/basesyntax/Application.java
maxline: this comments means that you need to create a constant with explanation naming instead of using `3` as hardcoded value. see the common mistakes please https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
```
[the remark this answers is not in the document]
maxline: comment not fixed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+    public Ball getRandomBall() {
+        Random random = new Random();
```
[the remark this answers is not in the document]
maxline: comment not fixed, move `random` to a class field please.

src/main/java/core/basesyntax/Lottery.java
maxline: not fixed. 
also  variable `value` is redundant, used only one times. so  `random.nextInt()` can be used directly in place where it is called

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        int value = random.nextInt(100);
+        Ball ball = new Ball(getRandomColor(), value);
+        return ball;
```
maxline, line 10: variable `ball` is redundant and can be inlined with `return` line.

maxline (changes requested):
Good job. But several comments from previous review ware not fixed. Please be more attentive with this in future.

sarakhmen (approved):
Good job :)

PR #879, opened 2023-04-20
Made a method for getting random balls

viprko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
     }
 }
+
```
sarakhmen, line 10: let's remove one redundant empty line here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].toString();
```
sarakhmen, line 7: Here, your random instance is created every time the method is called. random - is our dependency, let's initialize dependencies in the class fields. Also, as stated in common mistakes, it's better to use name() method in order to get string representation of the enums

sarakhmen (changes requested):
Let's improve your soluition :)

src/main/java/core/basesyntax/Application.java
viprko: Chekstyle won't approve my solution without an empty line at the end of the file

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = new Lottery().getRandomBall();
```
sarakhmen, line 6: you are still creating new lottery instances on each iteration. Let's create lottery instance only once before the loop. And also you can directly print the result of getRandomBall method call, so that no redundant variables are created

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final int COLORS_QTY = Colors.values().length;
```
sarakhmen, line 6: let's give it more clear name

sarakhmen (changes requested):
Overall good job, but let's improve your solution ;)

sarakhmen (approved):
Good job ;)

PR #880, opened 2023-04-20
Implements lottery task

m-pnsk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public String getRandomColor(Random random) {
+        return Color.values()[random.nextInt(Color.values().length)].toString();
```
sarakhmen, line 7: let's use name() method to get string representation of the enum, as stated in the common mistakes

sarakhmen (approved):
Overall good job, but you can improve your solution :)

src/main/java/core/basesyntax/ColorSupplier.java
m-pnsk: Forgot about it. Already fixed

PR #901, opened 2023-04-23
final solution version1.0

katrusyavshch:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return  new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(99));
```
olekskov, line 7: Consider using a single instance of `ColorSupplier` and `Random` by creating them as instance variables, rather than creating new instances every time the `getRandomBall()` method is called. This can improve performance and reduce unnecessary object creation. 🚀

src/main/java/core/basesyntax/ColoursDifferent.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum ColoursDifferent {
+    Pink,
```
olekskov, line 4: Enum values should be in uppercase, separated by underscores if needed. For example, PINK instead of Pink. 📚

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
olekskov, line 6: It's better to have a private static final `Random` instance instead of creating a new `Random` object every time `getRandomColor()` is called. 🌟

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,26 @@
+package core.basesyntax;
+
+public class Ball {
+    public String color;
```
olekskov, line 4: The 'color' field should be private to ensure proper encapsulation. Consider changing it to 'private String color;'. 🛠️

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
-        return null;
+        int index = new Random().nextInt(ColoursDifferent.values().length);
```
olekskov, line 7: Consider renaming the enum `ColoursDifferent` to `Color` for better readability and simplicity. 😊

src/main/java/core/basesyntax/Lottery.java
olekskov: Magic numbers should be constants with proper naming

olekskov (changes requested):

olekskov (approved):
Good job

PR #902, opened 2023-04-23
 create a simple lottery

DV7Art:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
sarakhmen, line 6: '3' could be considered as a magic number here. So, I'd recommend creating a constant for this number as described in common mistakes

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
...
+            System.out.println(i + 1 + ". " + new Lottery().getRandomBall().toString());
```
sarakhmen, line 7: Is it possible to not call toString method here?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(Colors.values().length);
```
sarakhmen, line 7: Random is your dependency here, and also an instance of this class is created each time the method is called. So, I'd recommend initializing 'random' in the class field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[randomIndex].toString();
```
sarakhmen, line 8: as described in common mistakes you should use name() method instead of toString to get string representation of enum

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int ballCount = 100;
```
sarakhmen, line 7: magic number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
...
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(ballCount));
```
sarakhmen, line 8: random and colorSupplier are also your dependencies here

sarakhmen (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random;
+
+    public ColorSupplier(Random random) {
+
```
rovein, line 9: Redundant empty line.

rovein (approved):
Goob job, but pay attention to the empty lines.

PR #906, opened 2023-04-24
Created classes Ball, Color, Lottery and enum ColorSupplier

margosha99:
In Lottery class method getRandomBall() returns random ball object with generated number and color in Main method three balls created with getRandomBall() and print info in the console

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,11 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery ball1 = new Lottery();
+        Lottery ball2 = new Lottery();
+        Lottery ball3 = new Lottery();
+        System.out.println(ball1.getRandomBall() + "\n"
+                + ball2.getRandomBall() + "\n"
+                + ball3.getRandomBall());
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        int value = random.nextInt(100);
+        ColorSupplier color = new ColorSupplier();
+        String randomColor = color.getRandomColor();
+        return new Ball(randomColor,value);
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball color is " + color + ". Ball number is " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public enum Color {
+    RED,
+    GREEN,
+    BLUE,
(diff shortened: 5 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
```
olekskov, line 6: Consider making the `random` variable `final` since it's not changed after initialization. 🛠

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Ball getRandomBall() {
+        int value = random.nextInt(100);
+        ColorSupplier color = new ColorSupplier();
```
olekskov, line 10: Consider making the `color` variable `final` since it's not changed after initialization. Also, it might be better to rename the variable to `colorSupplier` to avoid confusion with the actual color and move it to class variable as done for Random. 🎨

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery ball1 = new Lottery();
```
olekskov, line 6: You should create instances of the Ball class, not the Lottery class. Use the getRandomBall() method from the Lottery class to generate random balls. 🎱

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 7: It's better to create Random object as class variable in order to prevent creating of new instance each time method is called

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        int value = random.nextInt(100);
```
olekskov, line 9: magic numbers should be class constants with proper naming

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        ColorSupplier color = new ColorSupplier();
+        String randomColor = color.getRandomColor();
+        return new Ball(randomColor,value);
```
olekskov, line 12:
```suggestion
        return new Ball(randomColor, value);
```

olekskov (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball1 + "\n" + ball2 + "\n" + ball3);
```
okuzan, line 9: numbers in naming are often a sign you're doing something wrong, here you should use a for loop instead. Make sure for loop body is just one line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
okuzan, line 10:
```suggestion
        return Color.values()[index].name();
```
see [common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final Random random = new Random();
+    static final ColorSupplier colorSupplier = new ColorSupplier();
+    static final int MAX_NUMBER = 100;
```
okuzan, line 8: add access modifiers to all, leave static only for MAX_NUMBER constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+    public static Ball getRandomBall() {
```
okuzan, line 10: image

okuzan (changes requested):

uliana-andrushchuk (approved):

PR #913, opened 2023-04-30
added Ball, Color, Lottery classes; implemented methods from README task

Denis-Balako:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
+    private static final Random RANDOM = new Random();
+    private static final int RANDOM_COUNT = 100;
+
+    public static Ball getRandomBall() {
```
nacenik, line 10: [Don’t use static methods in your solution](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-use-static-methods-in-your-solution)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final Color color;
+    private final int number;
+
+    Ball(Color color, int number) {
```
nacenik, line 7: Don't forget about access modifiers

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(Lottery.getRandomBall());
```
nacenik, line 7: It would be best if you created a variable `lottery` and then reused it

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final ColorSupplier COLOR_SUPPLIER = new ColorSupplier();
```
nacenik, line 6: I'd recommend you do this variable non-static because static variables initialize only one time at the program's start.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball " + this.number + " with " + this.color.name().toLowerCase() + " color.";
```
nacenik, line 14: You don't need to use the keyword `this` cause you don't have variables with the same name here

nacenik (changes requested):
Good solution, let's improve it :)

KseniiaMakarova (approved):

PR #915, opened 2023-05-02
Hometask jv-lottery

MarynaKushniruk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
```
kozhukhovsky, line 7: remove empty line

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        int index = new Random().nextInt(Colors.values().length);
```
kozhukhovsky, line 8: image
It is necessary to move Random to a class variable, as in this example from the checklist

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        return Colors.values()[index].toString();
```
kozhukhovsky, line 9:
```suggestion
        return Colors.values()[index].name();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,34 @@
+package core.basesyntax;
+
+public class Ball extends Lottery {
```
kozhukhovsky, line 3:
```suggestion
public class Ball {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+
```
kozhukhovsky, line 6: remove empty line

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
```
kozhukhovsky, line 5:
```suggestion
public class Lottery {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Ball getRandomBall() {
+        Random random = new Random();
```
kozhukhovsky, line 8: It is also necessary to take out random. And like random, we also need to create a ColorSupplier object, not extend it.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
 
 public class Application {
     public static void main(String[] args) {
```
kozhukhovsky, line 4: It is enough to create one Lottery object, and then, in a loop with 3 iterations, get a new ball from it and pass it to the println method.
Also, when passing an object to println, there is no need to call their toString() method, it is called under the hood of the print method

kozhukhovsky (changes requested):

src/main/java/core/basesyntax/Application.java
MarynaKushniruk: All changes have been done.

olekskov (changes requested):

uliana-andrushchuk (approved):

PR #919, opened 2023-05-11
Hw 2 solution

Meysonsan:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        int number = new Random().nextInt(101);
```
uliana-andrushchuk, line 8: make ColorSupplier and Random() class level fields

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
uliana-andrushchuk, line 7: Please, make new Random() class level field
Do not forget `private` `final`

uliana-andrushchuk (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
Meysonsan: why Random() class level field should be done and private final Color[ ] colors = Color.value(); not needed, in which documentation can I look this up to avoid mistakes in the future?

Ivan95kos (approved):
Great job

PR #920, opened 2023-05-12
created classes(Ball,Color,Lottery)

krismva:
added implementation of methods

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        return String.valueOf(Colors.values()[new Random().nextInt(Colors.values().length)]);
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
+    GREEN, RED, BLACK, BLUE, YELLOW, WHITE
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(ColorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
+        return ball;
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,22 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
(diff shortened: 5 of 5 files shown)


Ivan95kos (changes requested):
make sure your build passes before requesting a review 😉

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: Consider renaming the enum to 'Color' instead of 'Colors' to follow naming conventions for singular names.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
olekskov, line 6: Consider changing the method getRandomColor() to non-static and creating an instance of ColorSupplier class to call this method.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(ColorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
olekskov, line 10: Consider creating a single instance of Random and reuse it, instead of creating a new Random instance each time you call getRandomBall().

Also magic numbers should be class constants with proper naming

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
```
olekskov, line 7: Consider changing the method from static to a non-static method, as it would be more appropriate for an object-oriented design.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        Ball ball = new Ball();
+        ball.setColor(ColorSupplier.getRandomColor());
```
olekskov, line 9: Create an instance of ColorSupplier and use its method to get a random color, instead of using a static method.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
olekskov, line 10: Please use cycle for repetitive actions

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
...
+        return String.valueOf(Colors.values()[new Random().nextInt(Colors.values().length)]);
```
olekskov, line 7: It's better practice to have only one Random instance for multiple calls. Consider creating a private final field Random instance.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
+        Ball ball = new Ball();
```
olekskov, line 8: I'd recommend to create all-arguments constructor for Ball class and use it instead of setters here. It will make code cleaner

olekskov (changes requested):

src/main/java/core/basesyntax/Colors.java
krismva: done

src/main/java/core/basesyntax/ColorSupplier.java
krismva: done

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        return String.valueOf(Color.values()[random.nextInt(Color.values().length)]);
```
aaananas, line 9:
```suggestion
        return Color.values()[random.nextInt(Color.values().length)].name();
```

aaananas (changes requested):

PR #921, opened 2023-05-14
add hw-2 solution

mykhailoKrlKrk:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,37 @@
...
+    //constructors
+    public Ball(String color, int number) {
+        this.number = number;
+        this.color = color;
+    }
+
+    public Ball() {
+    }
+
+    //getters and setters
```
okuzan, line 16: please remove the comments before comitting

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum name should be in singular

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
...
+    BLUE,
+    YELLOW,
+    RED,
+    PURPLE,
+    INDIGO,
+    GRAY,
+    WHITE
+
```
okuzan, line 11: remove the line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+            int index = new Random().nextInt(Colors.values().length);
+            return Colors.values()[index].name();
```
okuzan, line 8: check the indentation

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(new Random().nextInt(100));
+        ball.setColor(new ColorSupplier().getRandomColor());
+        System.out.print("Ball has a number: " + ball.getNumber() + " and color: " + ball.getColor());
+        return ball;
+        }
+    }
+
+
```
okuzan, line 15: remove the extra empty lines in the end, use IDE"s autoformatter before committing - indentation is off

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(new Random().nextInt(100));
+        ball.setColor(new ColorSupplier().getRandomColor());
+        System.out.print("Ball has a number: " + ball.getNumber() + " and color: " + ball.getColor());
```
okuzan, line 10: don't forget to remove System.out.print(ln)s before committing. We can see the result of how the code words in the main class (Application)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(new Random().nextInt(100));
```
okuzan, line 8: make 100 a class constant, don't forget to use CONSTANT_CASE and descriptive name

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+        ball.setColor(new ColorSupplier().getRandomColor());
```
okuzan, line 9: with this approach we create a new instance of ColorSupplier and Random per each method invocation. Let's make them class fields instead

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        System.out.println("Play a lottery!");
+        System.out.println("You get this balls:");
+
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
```
```suggestion
            System.out.println(lottery.getRandomBall());
```

okuzan: Create only one lottery instance, outside of the loop

okuzan (changes requested):
Let's make this solution better 😎

src/main/java/core/basesyntax/Lottery.java
mykhailoKrlKrk: can i use code refactor that macOs tools provide?Is it be good for code readablity  and standarts?
Знімок екрана 2023-05-14 о 18 12 23

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 7: Please move Random to class variable as done for Lottery class in order not to crate additional object every time method is called

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        System.out.println("Play a lottery!");
+        System.out.println("You get this balls:");
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
```
olekskov, line 11: We don't use ball variable anywhere else so it's creation is not necessary
```suggestion
            System.out.println(lottery.getRandomBall());
```

olekskov (changes requested):
good job overall, couple of moments left

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    Random random = new Random();
```
olekskov, line 6:
```suggestion
    private final Random random = new Random();
```

olekskov (approved):
Good job

PR #932, opened 2023-05-23
solution for jv-lottery task

vladis9:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+    private static final int LOTTERY_MAX_COUNT = 100;
+    private static final ColorSupplier supplier = new ColorSupplier();
```
okuzan, line 8:
```suggestion
    private static final int LOTTERY_MAX_COUNT = 100;
    private final Random random = new Random();
    private final ColorSupplier supplier = new ColorSupplier();
```
place fields of the same type together
think of Random and ColorSupplier as injected dependencies, it's more idiomatic to make them non-static

okuzan (changes requested):
Good job on you first attempt 💪

Ivan95kos (approved):
Great job

PR #934, opened 2023-05-27
lottery hometask done

happytroya:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball.toString());
```
olekskov, line 10: 1. method .toString() is called automatically for println() command, no need to explicitly set it:
```suggestion
            Ball ball = lottery.getRandomBall();
            System.out.println(ball);
```
2. There is no real benefit of introducing new variable to use it only once and right here:
```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball extends ColorSupplier {
```
olekskov, line 5: Ball shouldn't extend ColorSupplier.
It should be simple model with constructor/getters/setters required

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: Enum class should have name in singular form

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
```
olekskov, line 4: Lottery class should have Random and ColorSupplier as isntruments to create parameters of the ball

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
olekskov, line 7: please move initialization of new Random() to class varibale in order to avoid creation of new object each time method is called

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        Ball newBall = new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
olekskov, line 10: magic numbers should be constants with proper naming
```suggestion
        Ball newBall = new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_BALL_NUMBER));
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
olekskov, line 10:
```suggestion
        return Color.values()[index].name();
```
https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        return newBall;
```
olekskov, line 11: We should declare new variable if we plan to reuse it or when it's name adds clarity to complicated code.
When we have newBall = new Ball(...) it doesn't provide any benefits
```suggestion
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_BALL_NUMBER));
```

olekskov (changes requested):
Much better now, good job. Just some style comments left

olekskov (approved):
Good job!

hronograf (approved):

hronograf (approved):

PR #947, opened 2023-06-05
Hw 1 solution

zkolesnyk:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
okuzan, line 6: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-use-static-methods-in-your-solution

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        int randomNumber = new Random().nextInt(colors.length);
```
okuzan, line 8: same as with random in lottery class

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+        number = randomNumber;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", number=" + number
+                + '}';
+    }
```
okuzan, line 18: not a big issue, but toString is usually places after getters/setters

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: better to remove comments before committing

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
```
overpathz, line 6: That objects you won't change in the future should be final

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+
     public String getRandomColor() {
-        return null;
+        Color[] colors = Color.values();
```
overpathz, line 9: Instead of calling each time Color.values() to get Color array, you can create:
`public static Color[] colors = Color.values();`

Since enum is just static fields itself, they won't change at runtime during program, so you can use that to avoid unnecessary method calls and variables

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAXIMUM_NUMBER = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
overpathz, line 8: And make this object final.
Because as random object, it perfoms defined range of tasks (getting a color).
It has not an internal state which you will change. So from this point of view, this object is immutable.

(For instance, Person object can store surname as String. It won't be final, since we can change the surname. In case of colorSupplier, we do not want to do this. Also, final keyword can say java compiler to optimize some things under the hood)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public Ball getRandomBall() {
+        int randomNumber = random.nextInt(MAXIMUM_NUMBER);
```
overpathz, line 11: In this place, you store the random number.
But not store the random color and invoke the method right in Ball constructor.
Consider to do all in one style.

overpathz (changes requested):
Much better, keep it up!
Some changes and PR can be approved.

src/main/java/core/basesyntax/ColorSupplier.java
zkolesnyk: Variable 'colors' must be private and have accessor methods @from ide

src/main/java/core/basesyntax/Application.java
Ivan95kos: For the future, better to remove comments before committing

Ivan95kos (approved):
Great job, with minor comments

PR #948, opened 2023-06-06
added new classes and methods

JeniaSan:
added new classes Ball(with fields, get and set methods, and override toString methods) and Lottery(with method getRandomBall) , also added enum Color

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,37 @@
+
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball() {
+    }
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,26 @@
...
+        return color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + getColor() + ", " + "number: " + getNumber();
```
nacenik, line 23: you don't need to use getters here. You can just use class variables directly

nacenik (changes requested):
Great solution! Let's improve it a little bit :)

KseniiaMakarova (approved):

PR #954, opened 2023-06-12
Hw solution

AntonU91:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
okuzan, line 5: It's better to create a single Lottery instance and reuse it to get the random balls, instead of creating a new instance for each ball.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int randomNumber = new Random().nextInt(100);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
okuzan, line 8: Consider creating a single instance of ColorSupplier and reuse it, instead of creating a new instance each time the method is called.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int randomNumber = new Random().nextInt(100);
```
okuzan, line 7: Consider creating a single instance of Random and reuse it, instead of creating a new instance each time the method is called.

okuzan, line 7: 100 should be class constant, don't forget CONSTANT_CASE, access modifier

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
```
okuzan, line 7: delete the line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        int value = new Random().nextInt(Color.values().length);
```
okuzan, line 8: same for Random in this class - create a class field and reuse it

okuzan (changes requested):

PR #955, opened 2023-06-12
created Ball, Lottery and Colors(enum) classes

maxymmusiienko:
implemented getRandomColor() method in ColorSupplier


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Colors.values()[new Random().nextInt(Colors.values().length)].getColor();
```
okuzan, line 7: to improve the ColorSupplier class we can extract the creation of the Random object and store it as a field to avoid unnecessary object creation on every method call.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+        System.out.println(ball1.toString());
+        System.out.println(ball2.toString());
+        System.out.println(ball3.toString());
```
okuzan, line 11: Using numbered variables can make the code difficult to read and maintain. A for loop could simplify this code by allowing you to generate and print out any number of random balls with less repetitive code.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private String color;
+
+    Colors(String color) {
+        this.color = color;
+    }
+
+    public String getColor() {
+        return color;
+    }
```
okuzan, line 19: this is redundant, just call .name() on enum to get string representation

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
okuzan, line 7: 100 should be class constant, random, color supplier should be class fields

okuzan (changes requested):

overpathz (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
```suggestion
            System.out.println(lottery.getRandomBall());
        }
```

okuzan (approved):
Good job

PR #959, opened 2023-06-15
added solution

padu12:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random RANDOM = new Random();
+
     public String getRandomColor() {
```
andrii-hoienko, line 8: This methoid should return color, in our program color is enum, so please return propriate value here

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum name should better be singular

KseniiaMakarova (approved):

PR #963, opened 2023-06-16
Created class Lottery, methods to getRandomBall and getRandomColor

nastasiiaKulinich:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+        for (Ball ball : balls) {
+            ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
kozhukhovsky, line 12:
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

kozhukhovsky: It's not a good idea to run over an empty array, especially since when the ball is assigned, it doesn't get into the array, so it doesn't make sense. It is enough to do as I wrote

kozhukhovsky (approved):
Good job!

PR #971, opened 2023-06-24
Create Lottery and Ball class, Colors ENUM

AriohDK:
Create Lottery and Ball class, Colors ENUM
Implements method in ColorSupplier class


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

aaananas (approved):

PR #980, opened 2023-06-27
All tasks from README file are implemented

Alexsandr-Yablunovskyi:
'Lottery', 'Ball' classes and 'Color' ENUM are created. 'toString' method of the 'Ball' class is override. 'getRandomColor' and 'getRandomBall' methods are implemented.

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,37 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+        this.color = new ColorSupplier().getRandomColor();
+        this.number = new Random().nextInt(101);
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", number=" + number
+                + '}';
+    }
+}
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,5 @@
+package core.basesyntax;
+
+public enum Color {
+    YELLOW, BLUE, ORANGE, GREEN, BLACK, WHITE, PURPLE, RED, VIOLET, INDIGO
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Random;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+        this.color = new ColorSupplier().getRandomColor();
+        this.number = new Random().nextInt(101);
```
Elena-Bruyako, line 11: The maximum possible number should be 100

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return String.valueOf(Color.values()[new Random().nextInt(Color.values().length)]);
```
Elena-Bruyako, line 7: You can use static variable for Color.values().length to avoid calling length each time when getRandomColor() method executing.
Also, you can create inner variable 'int index' for better readability:
 int index = new Random().nextInt(arrayLength);

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Ball.java
Alexsandr-Yablunovskyi: Yes, the max possible number should be 100, that's why I put '101' in the nextInt() method, because to obtain a random integer from 0 to n (where n = 100, in our case) we shoult use nextInt(n+1). Or I've got it wrong?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,37 @@
...
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
```
liliia-ponomarenko, line 28: It's bad practice to leave code that you never used

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Color {
+    YELLOW, BLUE, ORANGE, GREEN, BLACK, WHITE, PURPLE, RED, VIOLET, INDIGO
```
liliia-ponomarenko, line 4: After each comma that follows an enum constant, a line break is optional but when expanding the list it looks more readable

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final int COLOR_QUANTITY = Color.values().length;
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int randomIndex = random.nextInt(COLOR_QUANTITY);
+        return String.valueOf(Color.values()[randomIndex]);
```
Elena-Bruyako, line 11: You could use name() method to get particular color

Elena-Bruyako (approved):

PR #981, opened 2023-06-27
implemented task

OleksandrSaven:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
+import java.util.Arrays;
+
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
KseniiaMakarova, line 12: let's create Lottery object before the loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(MAX_NUMBER));
```
KseniiaMakarova, line 12: we can declare new ColorSupplier() and new Random() as fields. so that we don't create a new object each time the method is called

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
nacenik, line 7: `Random` should be a class-level variable

nacenik (approved):

PR #992, opened 2023-06-29
was implemented methods for creating random balls

DmytroV95:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class Ball {
+
```
KseniiaMakarova, line 4: redundant empty line

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,39 @@
...
+    private String color;
+    private int number;
+
+    enum Color {
+        RED,
+        GREEN,
+        BLUE,
+        YELLOW,
+        ORANGE
+    }
```
KseniiaMakarova, line 14: please move it into separate class

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier randomColor = new ColorSupplier();
+        Ball randomlyBall = new Ball();
+        Random random = new Random();
+        for (int i = 0; i < random.nextInt(100); i++) {
+            randomlyBall.setColor(randomColor.getRandomColor());
+            randomlyBall.setNumber(randomColor.getRandomColor().length());
+        }
```
KseniiaMakarova, line 13: I don't get the logic of this loop
can we just calculate a random number and set it once to the ball?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier randomColor = new ColorSupplier();
+        Ball randomlyBall = new Ball();
+        Random random = new Random();
+        for (int i = 0; i < random.nextInt(100); i++) {
```
KseniiaMakarova, line 10: 100 should be a constant field

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,33 @@
...
+        this.randomColor = randomColor;
+    }
+
+    public Ball getRandomlyBall() {
+        return randomlyBall;
+    }
+
+    public void setRandomlyBall(Ball randomlyBall) {
+        this.randomlyBall = randomlyBall;
+    }
```
nacenik, line 24: Useless setters/getters

nacenik (changes requested):
Good solution, let's improve it)

src/main/java/core/basesyntax/BallColor.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum BallColor {
```
KseniiaMakarova, line 3:
```suggestion
public enum Color {
```
we can use this enum not only for balls

KseniiaMakarova (approved):

PR #993, opened 2023-06-29
implemented classes Ball, Lottery and ENUM ColorSupplier;

holsoni:
3 ball objects created and printed out in console

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,13 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
+        Ball secondBall = lottery.getRandomBall();
+        Ball thirdBall = lottery.getRandomBall();
+
+        System.out.println(firstBall);
+        System.out.println(secondBall);
+        System.out.println(thirdBall);
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", number=" + number
+                + '}';
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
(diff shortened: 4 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
```
andrii-hoienko, line 6: Please use loop to create objects

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
andrii-hoienko, line 4: There are color enum for color representation

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
andrii-hoienko, line 7: Overloaded expression, please introduce some variables. Also creation of Random object should in class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        String color = new ColorSupplier().getRandomColor();
```
andrii-hoienko, line 9: ColorSupplier object should be class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        int number = new Random().nextInt(MAX_NUMBER);
```
andrii-hoienko, line 10: number is not very good variable name, occures question what number? what is the purpose of it? etc

andrii-hoienko (changes requested):
Good, but need some impovements, also please read common miskates more cerafully

src/main/java/core/basesyntax/Lottery.java
holsoni: But what if we use it only once? Is it still bad practice?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
andrii-hoienko, line 6: This method should return color, in our program color is enum, so please return propriate value here

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
```
andrii-hoienko, line 6: this method can be simplified. Do you really need balls array and second loop?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        Color color = new ColorSupplier().getRandomColor();
```
andrii-hoienko, line 9: Please make ColorSupplier class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        int ballNumber = new Random().nextInt(MAX_BALL_NUMBER);
```
andrii-hoienko, line 10: Please make Random class level variable

andrii-hoienko (changes requested):
Still need some fixes. Hope you can manage everything fast!

KseniiaMakarova (approved):

PR #995, opened 2023-06-29
solution implemented

BlueVioletTeti:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
KseniiaMakarova, line 9: 100 should be a constant field

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+    private String randomColor = new ColorSupplier().getRandomColor();
```
nacenik, line 7: This is very bad practice. `ColorSupplier` should be a class-level variable and you should reuse it and call its method.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private int randomNumber = new Random().nextInt(MAX_BALL_NUMBER);
```
nacenik, line 8: This is very bad practice. `Random` should be a class-level variable and you should reuse it and call its methods.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
nacenik, line 7: Random should be a class-level variable

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        int numberOfBalls = 3;
+        Ball[] balls = new Ball[numberOfBalls];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < numberOfBalls; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
```
nacenik, line 11: Do you really need to create an array here? What about just using `System.out.println(lottery.getRandomBall());`?

nacenik (changes requested):

KseniiaMakarova (approved):

PR #997, opened 2023-06-29
Created method which create random balls

IhorKalaur:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+
+        System.out.println(ball1.toString());
+        System.out.println(ball2.toString());
+        System.out.println(ball3.toString());
```
KseniiaMakarova, line 11: please use loop

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(new Random().nextInt(100));
+        ball.setColor(new ColorSupplier().getRandomColor());
+        return ball;
```
KseniiaMakarova, line 10: 100 should be a constant field

also, we can move new Random() and new ColorSupplier() to the field level (declare them as fields), so that new objects are not created each time we call the method

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball;
+        for (int i = 0; i < 3; i++) {
+            ball = new Lottery().getRandomBall();
```
KseniiaMakarova, line 7: please create Lottery object outside of the loop

KseniiaMakarova, line 7:
```suggestion
            ball = new Lottery().getRandomBall();
```
```suggestion
            System.out.println(lottery.getRandomBall());
```

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
```
andrii-hoienko, line 8: This methoid should return color, in our program color is enum, so please return propriate value here

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
andrii-hoienko, line 4: There are color enum for color representation, please use it instead of stirng

andrii-hoienko (changes requested):
Good, but need there are few minor issues. Hope you can manage everything fast!

PR #999, opened 2023-06-29
Homework jv lottery

IvanFadieiev:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
KseniiaMakarova, line 8: use loop please

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(101));
```
KseniiaMakarova, line 7: 101 should be a constant field

also, we can declare new ColorSupplier() and new Random() on class level so that they are created only once

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #1003, opened 2023-06-29
created classes Lottery, Ball and Colors, implemented all methods

ArtemReizer:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int MAX_INDEX = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < MAX_INDEX; i++) {
+            Lottery lottery = new Lottery();
```
KseniiaMakarova, line 8: let's create Lottery object before the loop

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    private static final int BALL_NUM = 100;
```
KseniiaMakarova, line 6:
```suggestion
    private static final int MAX_BALL_NUMBER = 100;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int value = new Random().nextInt(BALL_NUM);
```
KseniiaMakarova, line 7: each time you call method, the value will be the same, because it is calculated only once
move calculation to method
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
```
nacenik, line 18: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

nacenik (approved):

PR #1004, opened 2023-06-29
Create class Ball and Lottery. Implements methods getRandomColor from…

OlegStashkiv:
… ColorSupplier, getRandomBall from Lottery.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -4,8 +4,8 @@ public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
         Lottery lottery = new Lottery();
-        System.out.println(lottery.getRandomBall());
-        System.out.println(lottery.getRandomBall());
-        System.out.println(lottery.getRandomBall());
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,9 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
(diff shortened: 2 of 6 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
KseniiaMakarova, line 9: use loop please

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        int number = new Random().nextInt(100);
+        return new Ball(colorSupplier.getRandomColor(), number);
```
KseniiaMakarova, line 9: 100 should be a constant field

KseniiaMakarova (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
nacenik, line 8: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
nacenik, line 7: `Random` should be a class-level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int bound = 100;
```
nacenik, line 6: Make it constant. [See here about constants and naming convention](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names)

nacenik (changes requested):
Good solution, let's improve it :)

KseniiaMakarova (approved):

PR #1023, opened 2023-07-07
added problem solution jv-lottery

Parabellum7:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        Color[] colors = Color.values();
+        Random random = new Random();
```
Ivan95kos, line 8: Random must be a class field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        String color = ColorSupplier.getRandomColor();
+        int number = (int) (Math.random() * 100) + 1;
```
Ivan95kos, line 6: Random and ColorSupplier must be a class field
use constants instead of numbers

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
Ivan95kos, line 6: Method must not be a static
```suggestion
    public String getRandomColor() {
```

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+        Ball ball1 = Lottery.getRandomBall();
```
Ivan95kos, line 5: Use one example of `Lottery` and `loop`

Ivan95kos (changes requested):

PR #1025, opened 2023-07-09
lottery_s_one

Lupen07:


okuzan (changes requested):
Please fix the build before requesting a review. Run `mvn clean package` command and make sure it passes. Go through the checklist mentioned in the task description, and review your code. Remove extra empty lines (after class definition for example). Don't make your methods or fields static. Use for loop in main. Make 100 a class constant. Random, ColorSupplier - class fields.

PR #1029, opened 2023-07-10
lottery test one

Lupen07:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
```
olekskov, line 7: Please do not use empty lines in the beginning of methods
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
+import java.util.Random;
+
+class ColorSupplier {
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        Color[] colors = Color.values();
+        if (colors.length == 0) {
+            throw new IllegalStateException("No colors defined");
+        }
```
olekskov, line 12: Cannot say it's mistake, it's definitely not, but I have never seen intentionally empty enums so this check seems excessive

olekskov (approved):
Good job, approved. Minor comments just for further consideration

PR #1031, opened 2023-07-13
implemented solution for lottery task

VolodymyrSkvira:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+}
+
+
```
Andriy-dz, line 9: exctra empty lines

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
Andriy-dz, line 7: let's create a class field for Random because you create it every time you create this method. If you make a class field, it will be created only once

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
...
+    RED,
+    BLUE,
+    YELLOW,
+    GREEN,
+    PURPLE;
+
+    @Override
+    public String toString() {
+        return name();
+    }
```
Andriy-dz, line 13: actually you don't need to override toString() rather use the name() method where you need to

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    public String getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        supplier.getRandomColor();
+        int random = new Random().nextInt(100);
+        return "Color: "
+                + supplier.getRandomColor().toString()
+                + ", "
+                + "Number: "
+                + String.valueOf(random);
```
Andriy-dz, line 14: let's make ColorSupplier and Rundom class fields. I also noticed that you call the getRandomColor method twice - this is not very correct. And as I previously pointed out, with enum it is better to use the name() method instead of toString(). Also, let's make the number 100 a constant

src/main/java/core/basesyntax/Ball.java
Andriy-dz, line 9: exctra empty lines

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
-public class Application {
+public class Application extends Lottery {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
liliia-ponomarenko, line 8: [Use for loop for creating several objects of the same class](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-for-loop-for-creating-several-objects-of-the-same-class)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
```
liliia-ponomarenko, line 6: redundant empty line

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private int index = new Random().nextInt(Colors.values().length);
```
liliia-ponomarenko, line 6: Don't make the index variable class-level (you should move it to method `getRandom()`. Only Random should make as class-level variable.
You can see examples in the common mistakes list:
image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
Use singular for class (enum) names

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int HUNDRED = 100;
```
liliia-ponomarenko, line 6: Which access modifier should you use here? Also, it's a bad name for constant better use MAX_NUMBER_OF_BALL

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int HUNDRED = 100;
+    private ColorSupplier supplier = new ColorSupplier();
+    private int random = new Random().nextInt(HUNDRED);
+
+    public String getRandomBall() {
```
liliia-ponomarenko, line 10: This method should return Ball but return String. For printing ball characteristics you should override method `toString() ` in class Ball

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int HUNDRED = 100;
+    private ColorSupplier supplier = new ColorSupplier();
+    private int random = new Random().nextInt(HUNDRED);
```
liliia-ponomarenko, line 8: the same issue. Only Random and ColorSupplier should make as class-level variables.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
-public class Application {
+public class Application extends Lottery {
```
liliia-ponomarenko, line 3: Why this class extends the Lottery?

liliia-ponomarenko (changes requested):
Let's fix your solution 😉

PR #1032, opened 2023-07-13
Hw 2

evgenkolner:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ballFirst = new Lottery().getRandomBall();
+        Ball ballSecond = new Lottery().getRandomBall();
+        Ball ballThird = new Lottery().getRandomBall();
+        System.out.println(ballFirst);
+        System.out.println(ballSecond);
+        System.out.println(ballThird);
```
Andriy-dz, line 10: can we write it in a loop so there is less code?)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                +
+                "number=" + number
+                +
+                ", color='" + color + '\''
+                +
+                '}';
```
Andriy-dz, line 20:
```suggestion
        return "Ball{"
                + "number=" + number
                + ", color='" + color + '\''
                + '}';
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
Andriy-dz, line 7: I recommend creating a field of the Random class, because you will create this object every time this method is called. And if you make it a class field, that will happen only once

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new Random().nextInt(100),new ColorSupplier().getRandomColor());
```
Andriy-dz, line 7: I also recommend making ColorSupplier and Random class fields. And 100 - its a magic namber. make it a constant

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public class Ball {
+    private int number;
+    private Color color;
```
Andriy-dz, line 5:
```suggestion
    private String color;
```

Andriy-dz (changes requested):
Good job! Lets improve your solution)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static Random random = new Random();
```
liliia-ponomarenko, line 6: You don't need to make this field static

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int quantity = 100;
+    private static ColorSupplier colorSupplier = new ColorSupplier();
+    private static Random random = new Random();
```
liliia-ponomarenko, line 8: the same here

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int quantity = 100;
```
liliia-ponomarenko, line 6: we use SNAKE_UPPER_CASE for constant names

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+    private static ColorSupplier colorSupplier = new ColorSupplier();
+    private static Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(random.nextInt(quantity),colorSupplier.getRandomColor().name());
```
liliia-ponomarenko, line 11: It's hard to read so I recommend creating some separate variables.
```suggestion
        return new Ball(random.nextInt(quantity), colorSupplier.getRandomColor().name());
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 5; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
liliia-ponomarenko, line 6: You create 5 objects of lottery, let's create only one ;)

liliia-ponomarenko (changes requested):
let's fix some mistakes ;)

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
 public class Application {
+
```
liliia-ponomarenko, line 4:
```suggestion
```

liliia-ponomarenko (approved):
Great job! ;)

PR #1035, opened 2023-07-15
Random balls

Parashchuk:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,18 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        //Create balls
```
liliia-ponomarenko, line 8: usually redundant comments it's bad practice

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,18 @@
...
+        Ball[] randomBalls = new Ball[3];
+        for (int i = 0; i < randomBalls.length; i++) {
+            randomBalls[i] = lottery.getRandomBall();
+        }
+
+        //Print all balls
+        for (Ball ball: randomBalls) {
+            System.out.println(ball.getColor());
+            System.out.println(ball.getNumber());
+        }
```
liliia-ponomarenko, line 18: Can you combine these loops?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
...
+        this.number = number;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public String getColor() {
+        return color;
+    }
```
liliia-ponomarenko, line 18: You can rewrite the method toString and remove these getters

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,23 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    enum Color {
```
liliia-ponomarenko, line 6: Why is Color nested class?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,23 @@
...
+        BLACK,
+        BLUE,
+        PINK,
+        PURPLE,
+        RED,
+        WHITE,
+        ORANGE
+    }
+
+    private static final int COLOR_COUNT = 9;
```
liliia-ponomarenko, line 18: this code is poorly supported, if you add a new color you will need to change this number. Let's remove it and use Color.values().length

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,23 @@
...
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(COLOR_COUNT)].name();
```
liliia-ponomarenko, line 21: Think about which variables should be local in the method and which should be class-level. Every time when your program call method getRandomColor new object of Random will be created

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
liliia-ponomarenko, line 7: the same here. also, it is hard to read so let's create some separate variables. 
Also, create a constant for the magic number

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            Ball randomBall = lottery.getRandomBall();
+
+            System.out.println(randomBall);
+        }
```
Andriy-dz, line 12: please, delete regular empty lines

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final int COLOR_COUNT = Color.values().length;
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+
```
Andriy-dz, line 10: same

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int maxAllowedNumber = 100;
```
Andriy-dz, line 6:
```suggestion
    private final int MAX_BALL_NUMBER = 100;
```

Andriy-dz (approved):
Good Job!

src/main/java/core/basesyntax/Application.java
Parashchuk: sorry for that, I usually just separate statements and function calls by logical blocks with empty lines

src/main/java/core/basesyntax/Lottery.java
Andriy-dz: I think you missed one keyword for constant

Parashchuk: Screenshot 2023-08-08 at 14 36 05

src/main/java/core/basesyntax/Application.java
Andriy-dz: this method is quite simple for readability, so I recommend not to add empty lines

src/main/java/core/basesyntax/Lottery.java
Parashchuk: > I think you missed one keyword for constant

iIdon't get it, I wrote the 'final' keyword, correct me if I am wrong, please

PR #1037, opened 2023-07-16
Jv lottery solution

ystankevych:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static Random random;
+
+    static {
+        random = new Random();
+    }
```
liliia-ponomarenko, line 10: why static? better initialize Random in the constructor

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public enum Colors {
```
liliia-ponomarenko, line 3: use singular for class (enum) names

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,16 @@
...
+    YELLOW,
+    GREEN,
+    BLUE,
+    NAVY,
+    PURPLE;
+
+    @Override
+    public String toString() {
+        return name().charAt(0) + name().substring(1).toLowerCase();
+    }
```
liliia-ponomarenko, line 15: you don't need to override the method toString(), just use name() or values (see example here https://howtodoinjava.com/java/enum/java-enum-string-example/)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final ColorSupplier COLOR_SUPPLIER;
+    private static final Random RANDOM;
+
+    static {
+        COLOR_SUPPLIER = new ColorSupplier();
+        RANDOM = new Random();
+    }
```
liliia-ponomarenko, line 12: it's bad practice to make instances (some service classes) constant.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public Ball getRandomBall() {
+        int number = RANDOM.nextInt(100);
```
liliia-ponomarenko, line 15: All magic numbers in your code should be constants. https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

liliia-ponomarenko (changes requested):
Good job but let's improve your solution )

liliia-ponomarenko (approved):
Great 🔥

PR #1042, opened 2023-07-24
created 5 classes w/ fields and method. As result by using method "getRandomBall" - complete the task.

tsuki99:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
-public class Application {
+public class Application extends Lottery {
     public static void main(String[] args) {
```
olekskov, line 4: Our main class shouldn't extend lottery, please create required object instead and use it to call required method further
```suggestion
public class Application {
    public static void main(String[] args) {
    Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: enum class name should be in singular form
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public static Ball getRandomBall() {
+        int randomNumber = new Random().nextInt(100);
+        Ball ball = new Ball(getRandomColor(), randomNumber);
+        return ball;
+    }
```
olekskov, line 10: 1. Lottery shouldn't extend ColorSupplier, it depends on it
2. Magic numbers should be extracted to constants with proper naming
3. Services we rely on are better as class variables in order not to create new object each time method is called
4. Do not use static methods unless it's really necessary
```suggestion
public class Lottery {
    private static final int MAX_BALL_NUMBER = 100;

    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

    public Ball getRandomBall() {
        int randomNumber = random.nextInt(MAX_BALL_NUMBER);
        Ball ball = new Ball(colorSupplier.getRandomColor(), randomNumber);
        return ball;
    }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Colors getRandomColor() {
+        int colorIndex = new Random().nextInt(Colors.values().length);
+        return Colors.values()[colorIndex];
     }
```
olekskov, line 9: Please rework according to comments left to Lottery class

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+public class Ball {
+    private Enum color;
+    private int number;
+
+    @Override
+    public String toString() {
+        return "color: " + color + "\n" + "number: " + number;
+    }
+
+    public Ball(Enum color, int number) {
```
olekskov, line 12:
```suggestion
    public Ball(Color color, int number) {
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private Enum color;
```
olekskov, line 4:
```suggestion
    private Color color;
```

olekskov (changes requested):

Ivan95kos (changes requested):
Please, fix all checks
Знімок екрана 2023-07-30 о 12 54 34

Ivan95kos (changes requested):
Good job overall!

okuzan (approved):
Well done!

PR #1048, opened 2023-07-25
homework is ready lottery

AndriiMarushchak:


Ball.java
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{" +
+                "color='" + color + '\'' +
+                ", number=" + number +
+                '}';
+    }
+
```
Ivan95kos, line 17: remove the last empty line in all classes/interfaces/methods where it is present
```suggestion
```

ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+import java.util.Random;
+
+class ColorSupplier {
+    private static final String[] colors = {"BLACK", "RED", "GREY", "YELLOW", "WHITE"};
```
Ivan95kos, line 4: Use `Enum Color` instead of constants

ColorSupplier.java
```diff
@@ -0,0 +1,10 @@
+import java.util.Random;
+
+class ColorSupplier {
+    private static final String[] colors = {"BLACK", "RED", "GREY", "YELLOW", "WHITE"};
+
+    public static String getRandomColor() {
+        Random random = new Random();
```
Ivan95kos, line 7: Random should be a class field

Lottery.java
```diff
@@ -0,0 +1,21 @@
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public static Ball getRandomBall() {
```
Ivan95kos, line 6: Method should not be a static
```suggestion
    public Ball getRandomBall() {

```

Lottery.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public static void main(String[] args) {
+        Ball ball1 = getRandomBall();
+        Ball ball2 = getRandomBall();
+        Ball ball3 = getRandomBall();
+
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
+    }
```
Ivan95kos, line 20: Create a `Main or Test` class and test your code there
Use a `for` loop to check instead of three objects

Ivan95kos (changes requested):

Test.java
```diff
@@ -0,0 +1,11 @@
+public class Test {
+    public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+        int numberOfBalls = 5;
+
+        for (int i = 0; i < numberOfBalls; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println("Ball " + (i + 1) + ": " + ball);
```
okuzan, line 8:
```suggestion
            System.out.println("Ball " + (i + 1) + ": " + lottery.getRandomBall());
```

ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+import java.util.Random;
+
+public enum Color {
+    RED, BLUE, GREEN, YELLOW, ORANGE;
```
okuzan, line 4: color enum should be in a separate file, this class should be ColorSupplier class

ColorSupplier.java
```diff
@@ -0,0 +1,11 @@
+import java.util.Random;
+
+public enum Color {
+    RED, BLUE, GREEN, YELLOW, ORANGE;
+
+    private static final Random random = new Random();
+
+    public static Color getRandomColor() {
```
okuzan, line 8:
```suggestion
    private final Random random = new Random();

    public Color getRandomColor() {

```
checklist issue, don't use static methods in this task

okuzan (changes requested):

Lottery.java
```diff
@@ -0,0 +1,12 @@
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private static final Random random = new Random();
```
okuzan, line 5:
```suggestion
    private final Random random = new Random();

```

ColorSupplier.java
```diff
@@ -0,0 +1,9 @@
+import java.util.Random;
+
+public class ColorSupplier {
+    public Color getRandomColor() {
+        return Color.getRandomColor();
+    }
+}
+
+
```
okuzan, line 9: remove these lines

Color.java
```diff
@@ -0,0 +1,8 @@
+import java.util.Random;
+
+public enum Color {
+    RED, BLUE, GREEN, YELLOW, ORANGE;
+        Color[] colors = Color.values();
+        return colors[random.nextInt(colors.length)];
+    }
+}
```
okuzan, line 8: this wouldn't even compile

okuzan (changes requested):
your files should be in this folder structure: 
image
now you have this:
image

Ivan95kos (approved):
Nice

PR #1054, opened 2023-07-26
Liliia hrebinyk hw lottery2

Mnemozine-Oh:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Ball();
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
+        }
```
okuzan, line 13:
```suggestion
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```
we don't need to save balls into array, let's print a ball immediately after it has been generated

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    public void setColor(Color color) {
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
```
okuzan, line 19:
```suggestion

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colSup = new ColorSupplier();
```
okuzan, line 8: avoid abbreviations, they decrease code readability and we don't gain anything from them. Coders spend way more time reading others' code than writing their own, so readability is more important than the time you spend typing a few more letters

okuzan (changes requested):
Good job overall!

okuzan (changes requested):
HW is tracked by platform!

olekskov (approved):

PR #1056, opened 2023-07-27
created lottery app

Oleksandr-Tymoshenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,19 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private enum Color {
```
liliia-ponomarenko, line 6: Why is Color nested class? use upper case for enums

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
Oleksandr-Tymoshenko: Used to do it nested from the other Pg language. What are the disadvantages of nested enum? I did this because this enum is only used in a method of one class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,22 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random RANDOM = new Random();
```
liliia-ponomarenko, line 6: Random and ColorSupplier  shouldn't be static just private final  (the same for Lottery class)
```suggestion
    private final Random RANDOM = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
liliia-ponomarenko: 1) Technically it can be nested but it is a generic list and can be used in any part of your application. Now you can leave it nested if you like, but in that case, you need to move it to the end of the class after the public methods
2) Now it is difficult to say why you had such a problem, maybe the code looked somehow different, but in general, these lists are constant values ​​and are named accordingly.  https://mate-academy.github.io/style-guides/java/java.html#s4.8.1-enum-classes 
If the code style check error occurs again, write about it in your private channel and we will solve the problem.

liliia-ponomarenko (changes requested):
let's fix a few mistakes 😉

liliia-ponomarenko (approved):
Great job! 🔥

PR #1057, opened 2023-07-28
tried to create app

PeresadaKostya:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(Colors randomColor, int i) {
+        this.color = color;
+        this.number = number;
+    }
```
olekskov, line 10: It will not work properly in current state as you accept variable "i" but trying to assign "number" and same for "randomColor" and "color" (but here also you have type mismatch)
```suggestion
    public Ball(String color, int number) {
        this.color = color;
        this.number = number;
    }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index];
```
olekskov, line 8: 1. Please extract new Random() as class variable not to create new object each time method us called
2. Method return type should be changed to work properly with constructor of Ball class
```suggestion
    private final Random random = new Random();

    public String getRandomColor() {
        int index = random.nextInt(Colors.values().length);
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: enum naming should be in singular form
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
olekskov, line 7: please extract ColorSupplier and Random to class variables similar to comment in ColorSupplier

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
Ivan95kos, line 10: All magic numbers, strings, symbols in your code should be constants.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball [] ball = new Ball[3];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < ball.length; i++) {
+            ball[i] = lottery.getRandomBall();
+            System.out.println(ball[i]);
+        }
```
Ivan95kos, line 10:
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < COUNT_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+    }
```
olekskov, line 11:
```suggestion
public class Lottery {
    private static final int MAX_BALL_NUMBER = 100;

    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

    public Ball getRandomBall() {
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_BALL_NUMBER));
    }
```

olekskov (approved):
Good job, approved
Please pay attention to comment below for further tasks

src/main/java/core/basesyntax/Lottery.java
PeresadaKostya: Thank you)

PR #1067, opened 2023-08-04
Erik Loskorikh done

lskErixon:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random rm = new Random();
+
     public String getRandomColor() {
-        return null;
+        Colors[] values = Colors.values();
+        int length = values.length;
+        int randIndex = new Random().nextInt(length);
```
sokimaaa, line 11: no need to create an instance of `class Random` each time #getRandomColor is being invoked. 

use the `rm` field that you have instead.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Lottery {
+    public void getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier l = new ColorSupplier();
```
sokimaaa, line 6: no need to create an instance of `ColorSupplier` each time #getRandomBall is being invoked. 

think about creating the class level field of ColorSupplier and using it instead.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Lottery {
+    public void getRandomBall() {
```
sokimaaa, line 4: let the Lottery#getRandomBall method returns the Ball, not a void.

The name of the method calls to us that we expect to `get` something, but we won't.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        Ball ball = new Ball();
+        ColorSupplier l = new ColorSupplier();
+        ball.setColor(l.getRandomColor());
+        ball.setNumber((int)(Math.random() * 100) + 1);
```
sokimaaa, line 8: 1. Let's use throughout our program the same methods of generating random value, e.g. `class Random` to make our program consistent and clear.

2. Please change all magic numbers with constant.

See https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery

sokimaaa (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,7 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lot = new Lottery();
```
aaananas, line 5: lets use full names, also your task was to create 3 balls

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,19 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random rm = new Random();
```
```suggestion
    private Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,19 @@
...
+        return rm.nextInt(rmInt) + 1;
+    }
+
     public String getRandomColor() {
-        return null;
+        Colors[] values = Colors.values();
+        int length = values.length;
+        int randIndex = rm.nextInt(length);
+        Colors colors = values[randIndex];
+        return colors.toString();
```
aaananas, line 17: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+enum Colors {
```
```suggestion
enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Lottery {
+    private Ball ball = new Ball();
```
aaananas, line 4: this should be a local variable, your method getRandomBall() is not creating a new ball - it changes properties of this ball, and return it
if you call this method a few times, stores those balls, then output it into console - you will see that they are all the same

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public String getRandomBall() {
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(colorSupplier.getRmInt(100));
```
aaananas, line 10: lets use Random object directly from here, ColorSupplier is meant to randomize colors not numbers
also 100 is a magic number ✨  use constants

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public class Lottery {
+    private Ball ball = new Ball();
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public String getRandomBall() {
```
```suggestion
    public Ball getRandomBall() {
```
you should return a ball from this method, not a string

aaananas (changes requested):

PR #1077, opened 2023-08-08
HW lottery

OlehReex:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball randomBall = new Ball();
+        randomBall.setColor();
```
Elena-Bruyako, line 6: Let`s impove your solution. 

First, when you use set method, you should provide argument to this method. 
Second, you can use constructor with variables to create a Ball object like this
new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_BALL_NUM));

Also. don't forget to [think about which variables should be local in the method and which should be class-level](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int randomIndex = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[randomIndex]);
```
Elena-Bruyako, line 8: you can use Color.values()[randomIndex].name() to the color

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,28 @@
...
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private int number;
+    private String color;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public void setColor(ColorSupplier color) {
```
Elena-Bruyako, line 16: Try to not leave methods that aren't used. As you change your logic and use constructor you don't need to use set methods anymore

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery ball = new Lottery();
```
Elena-Bruyako, line 6: Improve your naming, if you create Lottery object so the variable should also be lottery.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball {
+    private static final int FIGURE_COUNT = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
Elena-Bruyako, line 7: Remove unused variables

Elena-Bruyako (changes requested):
Good job!

Elena-Bruyako (approved):

PR #1081, opened 2023-08-08
Created a Ball class with the following fields: color and number and …

Eugene-Yukhimenko:
…replaced the toString() method, also created an enum with different colors.

  Implemented the getRandomColor() method in the ColorSupplier class, this method returns a random color based on enum values.
  Created a Lottery class with a getRandomBall() method that returns a ball, this method creates a ball with a random color and a random number, the maximum number being 100.
 Tested the program in the main() method.

### DIFF
--- a/src/main/java/ColorSupplier.java
+++ b/src/main/java/ColorSupplier.java
@@ -1,0 +1,8 @@
+import java.util.Random;
+
+public class ColorSupplier {
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index];
+    }
+}
--- a/src/main/java/ColorSupplier.java
+++ b/src/main/java/ColorSupplier.java
@@ -6,6 +6,5 @@ public class ColorSupplier {
     public Color getRandomColor() {
         int index = random.nextInt(Color.values().length);
         return Color.values()[index];
-
     }
 }
--- a/src/main/java/ColorSupplier.java
+++ b/src/main/java/ColorSupplier.java
@@ -2,6 +2,7 @@ import java.util.Random;
 
 public class ColorSupplier {
     private final Random random = new Random();
+
     public Color getRandomColor() {
         int index = random.nextInt(Color.values().length);
         return Color.values()[index];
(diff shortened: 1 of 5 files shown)


src/main/java/Lottery.java
```diff
@@ -0,0 +1,9 @@
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Random random = new Random();
+        int randomNumber = random.nextInt(100);
```
okuzan, line 6: Random and ColorSupplier should be class field, don't forget about appropriate access modifier
100 should be class constant, don't forget about CONSTANT_CASE

src/main/java/Main.java
```diff
@@ -0,0 +1,12 @@
+public class Main {
+    public static void main(String[] args) {
+        Lottery firstRandomBall = new Lottery();
+        Lottery secondRandomBall = new Lottery();
+        Lottery thirdRandomBall = new Lottery();
+
+        System.out.println(firstRandomBall.getRandomBall());
+        System.out.println(secondRandomBall.getRandomBall());
+        System.out.println(thirdRandomBall.getRandomBall());
+
```
okuzan, line 10: we need to create only one lottery instance, and then we call its method in a for loop

okuzan (changes requested):

src/main/java/ColorSupplier.java
```diff
@@ -0,0 +1,8 @@
+import java.util.Random;
+
+public class ColorSupplier {
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 5: Random should be class field same as in Lottery class in order not to create new object each time method is called

src/main/java/Lottery.java
```diff
@@ -0,0 +1,12 @@
+import java.util.Random;
+
+public class Lottery {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final int maxNumber = 100;
```
olekskov, line 6: constant are private static final and has UPPERCASE_NAMING
also it's better to place them before other variables
```suggestion
    private static final int MAX_NUMBER = 100;

    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

olekskov (changes requested):

olekskov (approved):
Good job

PR #1084, opened 2023-08-09
Feature/second branch

Cayman4010:
Created and tested Lottery application

### DIFF
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND_OF_NUMBER = 100;
+    Random random = new Random();
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return new Ball(colorSupplier.getRandomColor(), (random.nextInt(BOUND_OF_NUMBER) + 1));
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND_OF_NUMBER = 100;
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
liliia-ponomarenko, line 10: this variable also should make class-level

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        return new Ball(colorSupplier.getRandomColor(), (random.nextInt(BOUND_OF_NUMBER) + 1));
```
liliia-ponomarenko, line 11:
```suggestion
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(BOUND_OF_NUMBER) + 1);
```
these brackets are redundant. Also better create separate local variables for getting parameters and pass only names to the constructor. It's increase readability

liliia-ponomarenko (changes requested):
Good solution! Let’s improve it ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1085, opened 2023-08-10
created lottery application with three random balls

yevhenkukharchuk:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
-}
+package core.basesyntax;
+
+public class Application {
+    public static void main(String[] args) {
+        // create three balls using class Lottery and print information about them in console
+        Ball [] setOfBalls =
+                {new Lottery().getRandomBall(),
+                 new Lottery().getRandomBall(),
+                 new Lottery().getRandomBall()};
```
liliia-ponomarenko, line 9: One object of lottery enough

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+
+        for (Ball ball : setOfBalls) {
+            System.out.println(ball.toString());
```
liliia-ponomarenko, line 12:
```suggestion
            System.out.println(ball);
```
Method  `toString() ` will be called automatically

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+    private String color;
+    private int number;
+
+    public int getNumber() {
+        return number;
+    }
+
+    public String getColor() {
+        return color;
+    }
```
liliia-ponomarenko, line 13: better don’t  leave code that you never used

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
-        return null;
-    }
-}
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    public String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return "" + Color.values()[index];
```
liliia-ponomarenko, line 8:
```suggestion
        return Color.values()[index].name();
```
It’s  hardcode, just call method `name()`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball randomBall = new Ball();
+        ColorSupplier randomColor = new ColorSupplier();
+
+        randomBall.setColor(randomColor.getRandomColor());
+        randomBall.setNumber(new Random().nextInt(100));
```
liliia-ponomarenko, line 11: the same here
Also,[ all magic numbers in your code should be constants.](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants)

liliia-ponomarenko (changes requested):
Good start! Let’s improve your solution ;)
I don't know why github interpreted your changes like that, I hope it will be fixed in the next push

src/main/java/core/basesyntax/Application.java
Andriy-dz: not fixed

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
-    public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
-    }
-}
+package core.basesyntax;
+
+public class Application {
+    public static void main(String[] args) {
+        // create three balls using class Lottery and print information about them in console
+
```
Andriy-dz, line 6:
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBERS_BOUND = 100;
+    private Ball randomBall = new Ball();
```
Andriy-dz, line 7: do you really need this class field?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBERS_BOUND = 100;
+    private Ball randomBall = new Ball();
+    private Random randomNumber = new Random();
+    private ColorSupplier randomColor = new ColorSupplier();
```
Andriy-dz, line 9:
```suggestion
    private Random random = new Random();
    private ColorSupplier colorSupplier= new ColorSupplier();

```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Lottery.java
yevhenkukharchuk: Yes, you're right. It isn't neccesary.
For working three different Balls-creation in class Application i will relocate <Ball randomBall = new Ball()> to the method getRandomBall().
This statement <private static final int NUMBERS_BOUND = 100;> i used according to rules "All magic numbers in your code should be constants" 
Maybe it isn't neccesary to be private?

src/main/java/core/basesyntax/Application.java
yevhenkukharchuk: I try to do it in such way:

        Lottery lotteryBall = new Lottery();
        Ball [] setOfBalls =
                {lotteryBall.getRandomBall(), 
                 lotteryBall.getRandomBall(),
                 lotteryBall.getRandomBall()};

but I always got the same ball.
then i have to update now class Lottery: creation of Ball-object i should relocate in method getRandomBall()
I will make it now

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBERS_BOUND = 100;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
```
NazarSmith228, line 8: you may declare these fields as `final`, since you don't change them after initialization

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
-    public String getRandomColor() {
-        return null;
-    }
-}
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private Random randomIndex = new Random();
```
NazarSmith228, line 6: you may declare this field as `final` since you don't change it after initialization

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
-}
+package core.basesyntax;
+
+public class Application {
+    public static void main(String[] args) {
+        Lottery lotteryBall = new Lottery();
+        Ball [] setOfBalls =
+                {lotteryBall.getRandomBall(),
+                 lotteryBall.getRandomBall(),
+                 lotteryBall.getRandomBall()};
```
NazarSmith228, line 9: advice on how to improve this:

1. Declare a constant for the amount of balls: `private static final int NUMBER_OF_BALLS = 3`
2. You don't need an array since your goal is just to print information about the balls - therefore you can use a `for` loop and just print `lottery.getRandomBall()` to console

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
-public class Application {
-    public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
-    }
-}
+package core.basesyntax;
+
+public class Application {
+    public static void main(String[] args) {
+        Lottery lotteryBall = new Lottery();
```
NazarSmith228, line 5: this is not a `lotteryBall`, it's rather a `lottery` :)

NazarSmith228 (changes requested):
Great job. Now just fix the minor issues I mentioned in the comments and this PR will be approved :)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
-    public String getRandomColor() {
-        return null;
-    }
-}
+package core.basesyntax;
+
+import java.util.Random;
+
+public class ColorSupplier {
+    private static final Random randomIndex = new Random();
```
liliia-ponomarenko, line 6: Don’t  use static for instances 
```suggestion
    private final Random random = new Random();

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBERS_BOUND = 100;
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
```
liliia-ponomarenko, line 8:
```suggestion
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```
the same here

liliia-ponomarenko (changes requested):
You needed to make the instances final (not static). Let's fix it so your solution looks perfect ;)

liliia-ponomarenko (approved):
LGTM 🔥

PR #1086, opened 2023-08-10
Added solution for Lottery task

AlexRosenkreuz:
Created class Ball with fields color and number. Created enum Color with 10 colors. Created getRandomColor() and getRandomBall(). Printed info for 3 Ball objects using those methods.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class Application {
+    private static final int BALL_COUNT = 3;
+    private static Lottery lottery = new Lottery();
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for(int i = 1; i <= BALL_COUNT; i++) {
+            System.out.println(lottery.getRandomBall().toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,18 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String toString() {
+        return "Ball info:"
+                + "\n 1. Color: " + color
+                + "\n 2. Number: " + number
+                + "\n";
+    }
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER_BOUNDARY = 101;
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 
+import java.util.Random;
+
 public class Application {
+    private static final int BALL_COUNT = 3;
+    private static Lottery lottery = new Lottery();
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for(int i = 1; i <= BALL_COUNT; i++) {
```
liliia-ponomarenko, line 10:
```suggestion
        for(int i = 0; i < BALL_COUNT; i++) {
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String toString() {
```
liliia-ponomarenko, line 12:
```suggestion
    @Override
    public String toString() {
```
missed  annotation

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER_BOUNDARY = 101;
+    private static ColorSupplier colorSupplier = new ColorSupplier();
+    private static Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(NUMBER_BOUNDARY));
```
liliia-ponomarenko, line 11: better create separate local variables for getting parameters and pass only names to the constructor. It's increase readability

liliia-ponomarenko (changes requested):
Good start! Let’s improve your solution and fix check style mistakes ;)

src/main/java/core/basesyntax/Lottery.java
AlexRosenkreuz: Added variables

liliia-ponomarenko (changes requested):
Let’s fix check style mistakes ;)

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int BALL_COUNT = 3;
+    private static Lottery lottery = new Lottery();
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < BALL_COUNT; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
liliia-ponomarenko, line 9:
```suggestion
            System.out.println(lottery.getRandomBall());
```
this method will be called automatically

liliia-ponomarenko (approved):
Great job 🔥

src/main/java/core/basesyntax/Application.java
AlexRosenkreuz: Thanks, forgot about it)

PR #1087, opened 2023-08-10
Implemented desired methods

torbinofficial:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return String.valueOf(Color.values()[new Random().nextInt(Color.values().length)]);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private int maxNumber = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(maxNumber));
```
liliia-ponomarenko, line 9: the same here

liliia-ponomarenko (changes requested):
Good job! Let’s improve your solution!

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        return String.valueOf(Color.values()[random.nextInt(Color.values().length)]);
```
Andriy-dz, line 9:
```suggestion
        return Color.values()[random.nextInt(Color.values().length)].name();
```

Andriy-dz: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private static Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), random.nextInt(MAX_NUMBER));
```
Andriy-dz, line 10: ColorSupplier must also be made a class field

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #1088, opened 2023-08-10
Jv lottery hw

TetianaDenchyk:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
```
NazarSmith228, line 17: you don't need this getter method, since it is never used

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
NazarSmith228, line 6: you should better define a `Random` field in this class, so that you don't create a new instance of it every time this method is called
```suggestion
    private final Random random = new Random();
    
    public String getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
NazarSmith228, line 8: you should use `toLowerCase()` instead of `toString()` since `toString()` will just give you a string representation of the constant - `WHITE` for example, however you need `white`
```suggestion
        return Color.values()[index].toLowerCase();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall(Ball ball) {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
+
+        return ball;
```
NazarSmith228, line 11: two points:

1. you should define a constant field for `ColorSupplier` so that you don't create new instance of this class each time this method is called. same for `Random`
2. `100` is a magic number - you should create a dedicated contants for specifying the max ball number
3. this method should not receive any parameters - it should return "brand-new" ball as the result. by the way, passing objects as parameters and changing them inside of the method is an anti-pattern, since it leads to unexpected behavior. don't do that in future :)

```suggestion 
    private static final int MAX_BALL_NUMBER = 100;
    private final ColorSupplier colorSupplier = new ColorSupplier();
    private final Random random = new Random();
    
    public Ball getRandomBall() {
        Ball ball = new Ball():
        ball.setColor(colorSupplier.getRandomColor());
        ball.setNumber(random.nextInt(MAX_BALL_NUMBER));

        return ball;

```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Ball();
+            System.out.println(lottery.getRandomBall(balls[i]));
```
NazarSmith228, line 9: two points:

1. `3` is a magic number, you need to create a dedicated constant for the number of balls
2. you don't need to create an array for storing the balls, since your task is just to print them

```suggestion
        private static final int NUMBER_OF_BALLS = 3;
        // method declaration goes here
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
```

this change should be done after you address my [comment](https://github.com/mate-academy/jv-lottery/pull/1088/files#r1290240047)

NazarSmith228 (changes requested):
Good job! Now it's time to fix my remarks :)

liliia-ponomarenko (changes requested):
Firstly let’s fix check style mistakes ;)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(MAX_BALL_NUMBER));
+        return ball;
+    }
+
```
Andriy-dz, line 16: extra empty line

Andriy-dz (approved):

PR #1089, opened 2023-08-10
implement task

romazan07:
implement task

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,9 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,31 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color
+                + '\''
+                + ", number=" + number
+                + '}';
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
+    BLUE,
+    YELLOW,
+    BLACK,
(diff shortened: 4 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
+        return color;
+    }
```
NazarSmith228, line 9: you don't need this getter method, since it is not used anywhere

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
```
NazarSmith228, line 17: you don't need this getter method, since it is not used anywhere

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
NazarSmith228, line 6: you should better define a `Random` field in this class, so that you don't create a new instance of it every time this method is called
```suggestion
    private final Random random = new Random();
    
    public String getRandomColor() {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Lottery {
+    private Ball ball = new Ball();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
+        return ball;
+    }
```
NazarSmith228, line 13: three points:

1. you don't need this `ball` field, `getRandomBall()` method should return "brand-new" balls
2. you should better define a `Random` field in this class, so that you don't create a new instance of it every time this method is called
3. `100` is a magic number - you should create a dedicated contants for specifying the max ball number

```suggestion
    private static final int MAX_BALL_NUMBER = 100;
    private final ColorSupplier colorSupplier = new ColorSupplier();
    private final Random random = new Random();

    public Ball getRandomBall() {
        Ball ball = new Ball();
        ball.setColor(colorSupplier.getRandomColor());
        ball.setNumber(randim.nextInt(MAX_BALL_NUMBER ));
        return ball;
    }
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
NazarSmith228, line 7: `3` is a magic number, you need to create a dedicated constant for the number of balls - you can declare it as:
```
```

NazarSmith228 (changes requested):
Good job! Now it's time to fix my remarks :)

src/main/java/core/basesyntax/Application.java
liliia-ponomarenko: not fixed

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
```
liliia-ponomarenko, line 6: Redundant  empty line

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
liliia-ponomarenko, line 3: use singular  for class (enum) names

liliia-ponomarenko (changes requested):
Not all comments were fixed. And I added a couple of improvements ;)

NazarSmith228 (approved):
LGTM! :)

PR #1090, opened 2023-08-10
Solution for jv-lottery task!

pastcwdt:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        int numberOfBallsToPrint = 3;
```
liliia-ponomarenko, line 5: Let’s  make this variable constant

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Color.values()[index].toString();
```
liliia-ponomarenko, line 8: [Use name() for getting String representation of enum constants](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER = 100;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(new Random().nextInt(NUMBER), colorSupplier.getRandomColor());
```
liliia-ponomarenko, line 10: the same about Random

liliia-ponomarenko (changes requested):
Good job! Let’s  improve your solution ;)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        final int numberOfBallsToPrint = 3;
```
liliia-ponomarenko, line 5: You should  make  constant like:
private static final int NUMBER = 100; (Lottery::6)

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/Application.java
pastcwdt: got u!

Elena-Bruyako (approved):

PR #1091, opened 2023-08-11
homework

RomanYuzviak:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
NazarSmith228, line 6: this method should not be `static`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        int randomNumber = new Random().nextInt(100);
+        Ball randomBall = new Ball(ColorSupplier.getRandomColor(), randomNumber);
```
NazarSmith228, line 8: you should better define fields for `Random` and `ColorSupplier` in order not to create new instances of those every time this method is called

```
private final Random = new Random();
private final ColorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
NazarSmith228, line 6: this method should not be `static`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        int randomIndex = new Random().nextInt(Colors.values().length);
```
NazarSmith228, line 7: you should create a field for this `Random` instance, so that you don't create a new one each time this method is called:
```
private final Random random = new Random():
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Colors.values()[randomIndex].name();
```
NazarSmith228, line 8: you should add `.toLowerCase()` in order to get a better output

just `name()` will return you `WHITE` instead of `white`

```suggestion
        return Colors.values()[randomIndex].name().toLowerCase();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
```
NazarSmith228, line 15: you don't need this getter method

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
```
NazarSmith228, line 23: you don't need this getter method

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        for (Ball ball: balls) {
+            ball = Lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
NazarSmith228, line 10: you don't need an array to hold the `Ball` instances - your goal is just to print them

apart from that you need to declare a constant for the magic number `3`

don't forget that `getRandomBall()` should not be static, so you need to create an instance of `Lottery` class

after that you can use a `for` loop and just print the result of `getRandomBall()` method

```suggestion
        Lottery lottery = new Lottery():
        for (int i =0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

NazarSmith228 (changes requested):
Good job, now it's time to improve your solution :)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private final Random randomNumber = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+
+        Ball randomBall = new Ball(colorSupplier.getRandomColor(), randomNumber.nextInt(100));
```
NazarSmith228, line 11: you should declare a constant for the magic number `100` :)

something like `MAX_BALL_NUMBER`

NazarSmith228 (changes requested):
One last remark :)

src/main/java/core/basesyntax/Lottery.java
RomanYuzviak: image
don`t know how to handle it

RomanYuzviak: static keyword helped

NazarSmith228 (approved):
LGTM!

PR #1092, opened 2023-08-11
created classes and enum, implemented methods

validatorr:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
NazarSmith228, line 6: you should define a constant for the magic number `3`

something like:
```
```
```

File: src/main/java/core/basesyntax/Lottery.java [hunk 2]
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER = 100;
```
NazarSmith228, line 6: `NUMBER` is too general name

consider using something like `MAX_BALL_NUMBER`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        int number = random.nextInt(NUMBER);
+        String color = colorSupplier.getRandomColor().name();
```
NazarSmith228, line 12: you need to add `toLowerCase()` here, since `name()` returns you a String representation of the enum constant (`WHITE` instead of `white`)

```suggestion
        String color = colorSupplier.getRandomColor().name().toLowerCase();
```

PR #1093, opened 2023-08-11
create Ball class with 2 fields and overriding toString() method, cre…

vsych321:
…ate enum with possible colors, implement ,ethod in class ColorSupplier, where return a random color depended on enum values including Random class for it. Create Lottery class with its method, which return the ball, create a random ball and a random color and number(max = 100) and return ball. Test program in main, where create 3 balls using method from Lottery class and sout info

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int age;
+
+    public Ball(String color, int age) {
+        this.color = color;
+        this.age = age;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", age=" + age
+                + '}';
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int age;
```
NazarSmith228, line 5: this is not `age`, this is `number` :)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].name();
```
NazarSmith228, line 11: you need to call `toLowerCase()` here as well

`name()` returns you a String representation of enum constant (`WHITE` instead of `white`)

```suggestion
        return Color.values()[index].name().toLowerCase();
```

NazarSmith228 (changes requested):
Great job, just some minor issues to fix ;)

NazarSmith228 (approved):
LGTM! :)

PR #1094, opened 2023-08-11
add lottery with random color and random number

KDanil2000:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        int number = random.nextInt(BOUND);
+        String color = colorSupplier.getRandomColor().name();
```
NazarSmith228, line 12: you need to call `toLowerCase()` as well

`name()` will return you `WHITE` instead of `white`

```suggestion
        String color = colorSupplier.getRandomColor().name().toLowerCase();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
```
NazarSmith228, line 6: i would rename this to `MAX_BALLL_NUMBER`, since `BOUND` is not that informative

NazarSmith228 (changes requested):
Great job, just some minor issues to fix :)

NazarSmith228 (approved):
LGTM! :)

PR #1095, opened 2023-08-11
added solution

anastasiia-fi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery ball1 = new Lottery();
+        System.out.println(ball1.getRandomBall());
+        Lottery ball2 = new Lottery();
+        System.out.println(ball2.getRandomBall());
+        Lottery ball3 = new Lottery();
+        System.out.println(ball3.getRandomBall());
+
```
NazarSmith228, line 12: you don't need three instances of `Lottery` class - you can have just one instance and use a `for` loop

```suggestion
        Lottery lottery = new Lottery();
        for(int i = 0; i < NUMBER_OF_BALLS; i++ {
           System.out.println(lottery.getRandomBall());
        }

```
`NUMBER_OF_BALLS` - is a constant that you need to create:
```
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,31 @@
+package core.basesyntax;
+
+public class Ball {
+    public String getColour() {
+        return colour;
+    }
+
+    public int getNumber() {
+        return number;
+    }
```
NazarSmith228, line 10: you don't need getters here

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,31 @@
...
+    public void setColour(String colour) {
+        this.colour = colour;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    private String colour;
+    private int number;
```
NazarSmith228, line 21: you should place the fields before the methods:
```
private String colour;
private int number;

// methods go here...
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colour.values().length);
```
NazarSmith228, line 7: you need to declare a field in this class to hold an instance of `Random`:
```
```

this way you won't have to create new `Random` every time this method is called

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return (Colour.values()[index]).toString();
```
NazarSmith228, line 8: you need to add `toLowerCase()` here

`toString()` will return you `WHITE` instead of `white`

```suggestion
        return (Colour.values()[index]).toString().toLowerCase();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier randomColour = new ColorSupplier();
+        Random random = new Random();
```
NazarSmith228, line 9: you need to declare fields in this class to hold instances of `Random` and `ColorSupplier`:
```
```

this way you won't have to create new `Random` and `ColorSupplier` every time this method is called

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier randomColour = new ColorSupplier();
+        Random random = new Random();
+
+        ball.setColour(randomColour.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
NazarSmith228, line 12: you need to define a constant for `100`

something like `private static final int MAX_BALL_NUMBER = 100`;

NazarSmith228 (changes requested):
Good job, now it's time to fix my remarks :)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
```
Andriy-dz, line 6: you do not use this variable in methods. you also need to separate class fields and methods with an empty line. Likewise in other cases

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colour.values().length);
+        return Colour.values()[index].toString().toLowerCase();
```
Andriy-dz, line 9:
```suggestion
        return Colour.values()[index].name().toLowerCase();
```

Andriy-dz (changes requested):

src/main/java/core/basesyntax/Colour.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colour {
```
Elena-Bruyako, line 3: Try to avoid mistakes in English words in your code. Colour -> Color

Elena-Bruyako (changes requested):
Some comments need to be fixed

src/main/java/core/basesyntax/Colour.java
anastasiia-fi: image

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
Elena-Bruyako, line 9: Pay attention on the comment from Andriy

src/main/java/core/basesyntax/Colour.java
Elena-Bruyako: Agree with you. I wrote my comment because in other cases you use American word like ColorSupplier. Try to be consistent

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
anastasiia-fi: removed this variable ✔️

Elena-Bruyako: You don't need to remove this variable.
In your previous version you declare the field 

but not use this variable in method

int index = new Random().nextInt(Color.values().length);

So the correct version should be

```

}

```

Please look at common mistakes
[Think about which variables should be local in the method and which should be class-level](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
```
Elena-Bruyako, line 6: remove empty line
[Please don’t add redundant empty lines to your code.
](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#please-dont-add-redundant-empty-lines-to-your-code)
Please, read the common mistakes when you implement the task.

Elena-Bruyako (approved):

PR #1102, opened 2023-08-17
added all classes and methods

KhrystynaDidykh:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class Lottery {
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        String randomColor = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(100);
```
olekskov, line 13: Please move magic numbers to constants with proper naming

olekskov (changes requested):

PR #1104, opened 2023-08-18
Added solution

slaybrute:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,18 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Color[] COLORS = Color.values();
+    private static final int COLORS_AMOUNT = COLORS.length;
```
okuzan, line 7:
```suggestion
```
not worth it, you can get length from your first constant at any time with no cost

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[BALLS_AMOUNT];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
```
okuzan, line 14: you don't need to create and fill an array, you can print the result of the method call directly

okuzan (changes requested):

Ivan95kos (approved):
Nice

PR #1105, opened 2023-08-19
solution hw-2

Yaromyros:
First, I created 4 classes (Application - main, Ball, ColorSupplier, and Lottery). Also, I made an enum which stores my colors. 

Class ColorSupplier (in method getRandomColor), I added a random int value of my values in enum with command length. Then I built an object of color, that accepts a random index of color. Then I changed it to string and called the return statement.

In class Lottery, I constructed one constant for the magic number 100, in a method that returns a Ball object, I added a ball object which makes a ball with a random int number and string color, then I called a return statement for this object. 

In class Ball, I made two private valuables and constructors for objects class Ball. That I produced an override method toString which will print ball information.

The main class - application, contains a lottery object (class Lottery) which will produce a ball object. Then, when the lottery created a random ball, a new object called a random ball (class Ball) will take all functionality and values. Finally, we change this object to string type and print it into the console.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -4,9 +4,7 @@ public class Application {
     public static void main(String[] args) {
         Lottery lottery = new Lottery();
         for (int i = 0; i < 3; i++) {
-            Ball randomBall = lottery.getRandomBall();
-            String ballString = randomBall.toString();
-            System.out.println(ballString);
+            System.out.println(lottery.getRandomBall());
         }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball randomBall = lottery.getRandomBall();
+            String ballString = randomBall.toString();
+            System.out.println(ballString);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,10 +3,10 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball randomBall = lottery.getRandomBall();
+            String ballString = randomBall.toString();
+            System.out.println(ballString);
```
olekskov, line 9: System.out.println automatically calls .toString() method of object passed, no need to explicitly call it
```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int indexOfColor = new Random().nextInt(Color.values().length);
+        Color color = Color.values()[indexOfColor];
+        String getColor = color.toString();
+        return getColor;
```
olekskov, line 10: 1. Please make Random class variable in order not to create new object each time method is called
2. String representation of enum should be called with .name() method
```suggestion
        int indexOfColor = random.nextInt(Color.values().length);
        return Color.values()[indexOfColor].name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER_BOUND = 100;
+
+    public Ball getRandomBall() {
+        Ball newBall = new Ball(new Random().nextInt(NUMBER_BOUND),
+                new ColorSupplier().getRandomColor());
+        return newBall;
```
olekskov, line 11: Please make Random and ColorSupplier as class variable in order not to create new object each time method is called:
```suggestion
        return new Ball(random.nextInt(NUMBER_BOUND), colorSupplier.getRandomColor());
```

olekskov (changes requested):

PR #1107, opened 2023-08-20
homework - 2 - resolution

Yaromyros:
First, I created 4 classes (Application - main, Ball, ColorSupplier, and Lottery). Also, I made an enum which stores my colors.
Class ColorSupplier (in method getRandomColor), I added a random int value of my values in enum with command length. Then I built an object of color, that accepts a random index of color. Then I changed it to string and called the return statement.
In class Lottery, I constructed one constant for the magic number 100, in a method that returns a Ball object, I added a ball object which makes a ball with a random int number and string color, then I called a return statement for this object.
In class Ball, I made two private valuables and constructors for objects class Ball. That I produced an override method toString which will print ball information.
The main class - application, contains a lottery object (class Lottery) which will produce a ball object. Then, when the lottery created a random ball, a new object called a random ball (class Ball) will take all functionality and values. Finally, we change this object to string type and print it into the console.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,9 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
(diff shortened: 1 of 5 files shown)


kozhukhovsky (approved):
Good job!

PR #1109, opened 2023-08-23
Done HomeWork jv-lottery!

TkachukTara:


Rommelua (changes requested):
image

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lotteryObject = new Lottery();
+        Ball ball1 = lotteryObject.getRandomBall();
```
sokimaaa, line 6: Use for loop for creating several objects of the same class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        ListOfColors[] colors = ListOfColors.values();
+        int indexOfColor = new Random().nextInt(colors.length);
+        return colors[indexOfColor].toString();
+    }
+
+    public int getRandomINum() {
+        return new Random().nextInt(101);
```
sokimaaa, line 13: Don’t create several instances of the same class if you can use only one instance for your purposes

think about making a class-level field:
```
private Random random;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Lottery extends ColorSupplier {
```
sokimaaa, line 3: Using inheritance here is a bad idea. 

Use composition here instead.

sokimaaa (changes requested):

src/main/java/core/basesyntax/Application.java
TkachukTara: Done!

PR #1114, opened 2023-08-27
Solution for jv-lottery exercise

rpankiv:


Rommelua (changes requested):
image

rpankiv:
I have resolved all the errors from Maven. Please review it again. Thanks!

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
```
sokimaaa, line 7: use loop statement to avoid repeating code.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
sokimaaa, line 7: with current implementation you instantiate `class Random` each time method is invoked.

you should instantiate this class one time and use it multiple times.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
sokimaaa, line 3: `enum Color`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
sokimaaa, line 6: make method non-static

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,21 @@
...
+        return randomNumber;
+    }
+
+    public static void setRandomNumber(Random randomNumber) {
+        Lottery.randomNumber = randomNumber;
+    }
+
+    public static Ball getRandomBall() {
+        Ball rundomBall;
+        rundomBall = new Ball(ColorSupplier.getRandomColor(), randomNumber.nextInt(100));
```
sokimaaa, line 18: avoid using magic number

sokimaaa (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
rpankiv: Could you please clarify what you mean? I implemented this based on the examples provided in the theory section of the OOP module on the Mate Academy platform. This is the approach I was taught by your platform.

I am certainly open to acquiring new knowledge and would greatly appreciate it if you could provide me with an example of a different implementation.
Screenshot 2023-09-02 at 20 44 28

src/main/java/core/basesyntax/Colors.java
rpankiv: Could you please clarify your comment? Is it considered bad practice to use plural to name entities in Java?

src/main/java/core/basesyntax/ColorSupplier.java
rpankiv: I can make that change, but I would appreciate it if you could explain to me why it should be changed. What are the potential problems that could occur if we use a static method here?

Thanks in advance.

src/main/java/core/basesyntax/Lottery.java
rpankiv: Could you please explain what you mean? The only number here is 100; what is so magic about it? What do you want to change?

src/main/java/core/basesyntax/Application.java
rpankiv: Got it, thank you. I'll make the update.

PR #1116, opened 2023-08-27
Hw 2 solution

ereshetnik:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color: " + color + ", number: " + number;
+    }
+
```
okuzan, line 34:
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].toString();
```
okuzan, line 8: random should be a class field
Also, a checklist issue:
image

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball(); // color and number are created in the default constructor
+        return ball;
```
okuzan, line 6: this defeats the purpose of the task, Lottery class should be responsible for generating a ball, what if we want to create a plain ball, without generating random color and number? Your implementation doesn't allow this

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
```
okuzan, line 6: toString is called by default when we put an object in System.out.println
We should create only one lottery instance, right now you are doing it three times, per each iteration of the loop

okuzan (changes requested):

PR #1119, opened 2023-08-31
Hw 2 lottery solution

GoldHolub:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public void getRandomBall(Ball ball) {
```
aaananas, line 6:
```suggestion
    public void getRandomBall() {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
...
+        ball.setColor(new ColorSupplier().getRandomColor());
+        ball.setNumber(new Random().nextInt(100));
```
aaananas, line 8: lets make ColorSupplier and Random class fields, and constant for max number

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int indexOfColor = new Random().nextInt(Colors.values().length);
+        return Colors.values()[indexOfColor].toString();
```
aaananas, line 8: see common mistakes

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int indexOfColor = new Random().nextInt(Colors.values().length);
```
aaananas, line 7: lets make Random calss field - to avoid creating a new instance each time this method is called

aaananas (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxNumber = 100;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private Ball ball = new Ball();
+
+    public Ball getRandomBall() {
```
sokimaaa, line 11: with the current implementation getRandomBall dont generate a new ball, just continuously update existing one 

so, Ball ball sholdn't be class-level field

sokimaaa (changes requested):

src/main/java/core/basesyntax/Lottery.java
GoldHolub: Understood and fixed. But I don't understand one moment - in this case I create 4 objects (One obj in class Lottery and 3 of them in Application) getRandomBall have only one obj just continuously update existing one and then transfer the result in another 3 objects (balls I created in Application class). But if ball shouldn't be class-level field then I create 6 objects instead of 4 and have the same result in final. So basically I do not understand why we are doing something like this. Thanks.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
 public class Application {
+    private static final int NUMBER_OF_OBJECTS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_OBJECTS];
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < NUMBER_OF_OBJECTS; i++) {
+            balls[i] = new Ball();
```
sokimaaa, line 11: `balls[i] = new Ball()` line has no sense because method Lottery#getRandomBall rewrites var with new value

src/main/java/core/basesyntax/Lottery.java
sokimaaa: no need to create a ball object in the Application because you have the #getRandomBall method for this. 

see comment above.

sokimaaa (changes requested):

aaananas (approved):

PR #1120, opened 2023-09-01
Create Ball class and override toString() method.

ipantazi:
Create enum with different colors.
Implement method getRandomColor() in class ColorSupplier. 
Create a Lottery class with getRandomBall() method. 
Create three balls in loop and print information about them in the console.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,9 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
kozhukhovsky, line 7:
```suggestion
            System.out.println(lottery.getRandomBall());
```

toString on an object is called under the hood inside the println method

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

kozhukhovsky (approved):
Good job!
There is a mistake with enum naming, but just don't make it in the future.
All classes are named in the singular.

PR #1121, opened 2023-09-01
Hw lottery solution

Hryniv:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
...
+import core.basesyntax.models.Ball;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Lottery lottery = new Lottery(colorSupplier);
+
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
```
sokimaaa, line 11: use loop statements if you got repeated code.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+import core.basesyntax.models.Color;
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final int COLOR_COUNT = 8;
+
+    public Color getRandomColor() {
+        return Color.values()[new Random().nextInt(COLOR_COUNT)];
```
sokimaaa, line 10: you instantiate `class Random` every time `getRandomColor` method is invoked.

Should be instantiated once and reused multiple times. (think about making it class-level field)

sokimaaa (changes requested):

sokimaaa (approved):

PR #1123, opened 2023-09-04
create class Lottery and implement methods "getRandomBall" and "getRa…

dima-semeniuk:
…ndomColor"

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
...
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
d1sam, line 10: The toString method is automatically called when something is printed using println. Don’t use toString() or String.valueOf()(it will call toString() under the hood).

d1sam (changes requested):
You're close, but a one revision is necessary.

d1sam (approved):
Great job on fixing the code!

PR #1127, opened 2023-09-07
add classes Ball, Lottery

VoitseshukV:
I changed the type of ColorSupplier.getRandomColor() method to enum Color (from String)

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,8 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index];
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall().toString());
+        }
```
olekskov, line 7: 1. Please extract creation of lottery object from cycle, we may obtain required result creating only 1 object but not new for each cycle iteration
2.  System.out.println method automatically calls .toString() method of object passed, no need to explicitly indicate it
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < 3; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        int number = new Random().nextInt(99) + 1;
+        return new Ball(new ColorSupplier().getRandomColor(), number);
+    }
+
+}
```
olekskov, line 12: 1. Magic numbers should be constants with proper naming
2. Services we rely on should be class variables in order not to create new object each time method is called
3. Pls remove unnecessary empty line
```suggestion
public class Lottery {
    private static final int NUMBER_LIMIT = 100;
    
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

    public Ball getRandomBall() {
        return new Ball(colorSupplier.getRandomColor(), random.nextInt(NUMBER_LIMIT));
    }
}
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 7: Please extract Random as class variable as advised in previous comment

olekskov (changes requested):

Ivan95kos (approved):
Nice

PR #1128, opened 2023-09-09
created all classes with methods

flekros:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,13 @@
...
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball ball = lottery.getRandomBall();
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        System.out.println(ball);
+        System.out.println(ball1);
+        System.out.println(ball2);
+
```
d1sam, line 13: Use for loop for creating several objects of the same class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
d1sam, line 8: Think about which variables should be local in the method and which should be class-level

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+        String color = colorSupplier.getRandomColor().toString();
+        int number = getRandomNumber();
+        Ball ball = new Ball();
+        ball.setColor(color);
+        ball.setNumber(number);
+        return ball;
+    }
+
+    public int getRandomNumber() {
+        return new Random().nextInt(100);
```
d1sam, line 18: Think about which variables should be local in the method and which should be class-level

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String color = colorSupplier.getRandomColor().toString();
+        int number = getRandomNumber();
+        Ball ball = new Ball();
+        ball.setColor(color);
+        ball.setNumber(number);
+        return ball;
+    }
+
+    public int getRandomNumber() {
```
d1sam, line 17: Think about which access modifier should be on this method

d1sam (changes requested):
You're on the right track, but some revisions are needed.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        String color1 = colorSupplier.getRandomColor().toString();
+        int number1 = getRandomNumber();
+        return new Ball(color1, number1);
+    }
+
+    private int getRandomNumber() {
+        return random.nextInt(100);
```
d1sam, line 16: Check [this article](https://www.geeksforgeeks.org/java-util-random-nextint-java/) about `random.nextInt()` and think about upper limit.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        String color1 = colorSupplier.getRandomColor().toString();
+        int number1 = getRandomNumber();
```
d1sam, line 11: Think about better names for these variables. Do you need digit `1` at the end?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,42 @@
...
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public Ball() {
+
+    }
```
d1sam, line 14: Do you need a default constructor?

d1sam (changes requested):
Great job on fixing the code! But you need to check a few things in your code.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_REPEAT = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
bhdnchui, line 7: Please, remove that unnecessary doc.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+        Lottery lottery = new Lottery();
+        for (int i = 0;i < NUMBER_OF_REPEAT;i++) {
```
bhdnchui, line 9: Please, format your code.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
```
bhdnchui, line 11: Don't you think the `ball` variable is redundant here?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='"
+                + color
+                + '\''
+                + ", number="
+                + number
+                + '}';
```
bhdnchui, line 36: This formatting is redundant. I think that would be preferable:
```
 return "Ball {"
                + "color='" + color + '\''
                + ", number=" + number
                + '}';
```

src/main/java/core/basesyntax/Application.java
bhdnchui: You can use `Ctrl + Alt + L` or `⌘⌥L` (depending on your OS) to automatically format your code while you are using Intellij Idea. [Here](https://resources.jetbrains.com/storage/products/intellij-idea/docs/IntelliJIDEA_ReferenceCard.pdf) are a lot of other shortcuts.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor().toString();
+        int number = getRandomNumber();
+        return new Ball(color, number);
+    }
+
+    private int getRandomNumber() {
+        return random.nextInt(101);
```
bhdnchui, line 16: Please, follow [this](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants) rule.

bhdnchui: It would be good to put the maximum ball number in a separate variable and then add `1` to the random number to get the expected result.
```
....
return random.nextInt(MAX_NUMBER) + 1;
```
In this way, the number 100 will be stored in a good format and the question `Why it is 101 and not 100?` will not appear.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public enum Colors {
```
bhdnchui, line 3: It would be better to rename that enum. 
`Colors` -> `Color`.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,38 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
bhdnchui, line 4: Please, store a `Color` enum value instead of a string literal here.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor().toString();
```
bhdnchui, line 10: When you start to store a `Color`enum value in `Ball` class it will be redundant to call `toString` method and your will be able to simplify your code in that way`Color color = colorSupplier.getRandomColor()`.
Isn't that wonderful?

src/main/java/core/basesyntax/Ball.java
bhdnchui: Also, make sure that you will process enum value instead of a string literal value across your solution.

bhdnchui (changes requested):
Please, take a look at my comments.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Random random = new Random();
+    Color [] colors = Color.values();
```
bhdnchui, line 7: Please, add an access modifier here.

bhdnchui: Also, please note that your solution did not pass automatic verification. This is a check on the GitHub platform that validates your code for some errors. If you read the logs, you can see that she pointed out this mistake of yours.
image
image

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+private static final int NUMBER_OF_REPEAT = 3;
```
bhdnchui, line 4: Please, format your code correctly.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+    private static final int MAX = 100;
```
bhdnchui, line 8: It is a good practice to declare constants first. Please, place constant declaration before regular class fields.
Also, think about more descriptive name. Maybe, `MAX_BALL_NUMBER` would more descriptive?

bhdnchui (changes requested):
You did well. Only some small edits are needed.

d1sam (approved):
Good job on fixing the code.

PR #1135, opened 2023-09-11
jv-lottery-hw

TarasYurkovskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0;i < 3;i++) {
+            System.out.println(lottery.getRandomBall().toString());
+        }
```
d1sam, line 9: All magic numbers in your code should be constants.
Make code prettifier, add spaces after ;
The toString method is automatically called when something is printed using println. You don't need to call this method in `System.out.println`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Color color = Color.values()[index];
+        return color.toString();
```
d1sam, line 9: `toString()` is common for all enum constants. It means, that for every constant `toString()` will be returning default, that’s not ok. So it’s better to use standard method of enum `name()` that will be returning always String representation of concrete enum constant.

d1sam (changes requested):
Some changes should be done according my comments. Please pay attention to details.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
```
kshuryhin, line 4: I would say making `color` variable of type `Color` is beneficial from readability perspective.

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
kshuryhin, line 7: Still valid point about magic numbers. Please, create a constant which will represent number of balls.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER_OF_BOUND = 100;
```
kshuryhin, line 6: Please, read about `nextInt(int bound) `upper bounds. No, maximum possible value from random generating will be 99.
[Java doc](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt-int-)

kshuryhin (changes requested):
Nice try, but there are some points to consider

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public Ball(Color color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
```
d1sam, line 15: if you use Color enum as a field of Ball class you should be very careful. Use name() for getting String representation of enum constants

d1sam (changes requested):

d1sam (approved):
Well done.

PR #1136, opened 2023-09-12
implemented Lottery and Ball classes.

hehetenya:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
dshulha, line 6: All magic numbers in your code should be constants.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
```
dshulha, line 7: Think about which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random()
```
dshulha, line 9: Think about which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
```
dshulha, line 6: Think about upper limit of ball number. Please see [this article](https://www.geeksforgeeks.org/java-util-random-nextint-java/) to learn about upper bound of `random.nextInt()`

dshulha (changes requested):
Good work, but you need to make some review changes. Pay attention to class variables please and use them, where is needed. Also learn about upper bound of `random.nextInt()`. I believe, you will do the best.

dshulha (changes requested):
One more try to reject to show this on mate platform.

d1sam (changes requested):
One more try to reject to see this on mate platform.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
```
kshuryhin, line 4: I would say, making color variable of type `Color` would be beneficial from readability perspective, but that is minor.

PR #1137, opened 2023-09-13
jv-lottery

rafik2211:
create Ball class with the following fields: color and number and override toString() method. create enum with different colors. create a Lottery class with getRandomBall() method, which will return the ball. This method should create a ball with a random color and a random number and return this ball from the method.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,8 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(Lottery.getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,8 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+
         for (int i = 0; i < 3; i++) {
-            System.out.println(Lottery.getRandomBall());
+            System.out.println(lottery.getRandomBall());
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(Lottery.getRandomBall());
```
Ivan95kos, line 6: create an instance of `Lottery`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        Color randomColor = ColorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(100) + 1; // Random number between 1 and 100
+        return new Ball(randomColor, randomNumber);
+    }
```
Ivan95kos, line 10: ColorSupplier and Random must be class fields
The method must not be a static
Remove the comments, code should be understandable without them

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Random random = new Random();
+
+    public static Color getRandomColor() {
```
Ivan95kos, line 8: The method must not be a static
```suggestion
    public Color getRandomColor() {
```

Ivan95kos (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+    private final Random random;
+
+    public Lottery() {
+        this.colorSupplier = new ColorSupplier();
+        this.random = new Random();
+    }
+
+    public Ball getRandomBall() {
+        Color randomColor = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(100) + 1;
```
kozhukhovsky, line 16: better, if 100 move to the class constant, because its magic number

kozhukhovsky (approved):

PR #1141, opened 2023-09-17
added new classes and implemented new methods

fedorovychh:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery ball1 = new Lottery();
+        Lottery ball2 = new Lottery();
+        Lottery ball3 = new Lottery();
+        Lottery[] balls = {ball1, ball2, ball3};
+        for (Lottery ball : balls) {
+            System.out.println(ball.getRandomBall());
+        }
```
mnyshenko, line 11: A few points about this part:
1) Variables names should reflect their content. You have three "ball" variables, that contain lottery objects, this doesn't seem right.
2) You need only 1 lottery object, to create any number of balls, because the lottery class has a particular method for this, which makes a new ball every time you call that method (getRandomBall())

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public void setColor(String color) {
+        this.color = color;
+    }
```
mnyshenko, line 9: It will be more convenient to create a constructor for this class, which will let you create and fulfill an instance of this class using only 1 statement instead of 3.

Ball ball = new Ball();
ball.setNumber(smth);
ball.setColor(smth);

vs 

Ball ball = new Ball(smth, smth);

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
mnyshenko, line 4: You have a specific enum to represent colors, please use it here instead of the String.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index];
```
mnyshenko, line 8: A few points about this part:
1) Please pay attention to this part in the common mistakes document: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level
Because now you create a Random instance every time you call getRandomColor(), this is insufficient.
2) You call Color.values() twice, these parts should be extracted to a separate variable.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int value = random.nextInt(100);
+        ColorSupplier color = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setNumber(value);
+        ball.setColor(color.getRandomColor().toString());
+        return ball;
```
mnyshenko, line 13: 1) Again, pay attention to https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level
2) One more point about using magic numbers in code https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants
3) When you create a constructor for a Ball class, don't forget about this point https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-temporary-variables-if-you-can-directly-return-value-from-the-method

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random random = new Random();
+    private static final Color[] values = Color.values();
```
mnyshenko, line 7: pay attention to variable namings, imagine that you have a class with hundreds of lines, it will take some time to figure out what values actually are. The better naming will be colors

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
```
mnyshenko, line 4: I still recommend you to make type of this variable as Color, not a String

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private static final Random random = new Random();
+    private static final ColorSupplier color = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        int value = random.nextInt(MAX_NUMBER);
```
mnyshenko, line 11: Please, pay attention to what you pass to this method and how it actually works. Now the maximum possible value, that this method can return, is 99.
image

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private static final Random random = new Random();
+    private static final ColorSupplier color = new ColorSupplier();
```
mnyshenko, line 8: Please, choose a more appropriate naming, because this variable doesn't hold **color**, it holds **colorSupplier**

mnyshenko (changes requested):

d1sam (approved):
Well done!

PR #1143, opened 2023-09-17
First solution of jv-lottery

QbaSekowski:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
```
Rommelua, line 10: image

QbaSekowski: Done. Now only one ball is created in Main class.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball = new Ball();
```
Rommelua, line 8:
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            ball = new Lottery().getRandomBall();
```
Rommelua, line 10: Do not create a Lottery inside the loop. You need 3 balls. Creating 3 Lotteries here is like visiting 3 different markets when you need to buy 3 milk bottles.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int LIMIT_OF_NUMBERS = 101;
+
+    public Ball getRandomBall() {
+        Random randomValuesSupplier = new Random();
+        int randomInt = randomValuesSupplier.nextInt(LIMIT_OF_NUMBERS);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Rommelua, line 11: image

Rommelua (changes requested):

QbaSekowski:
- created 3 balls, but still not sure if that's how it should be :)
- changed the name of enum
- changed two local variables to class-level

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
Rommelua, line 13: image

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int LIMIT_OF_NUMBERS = 101;
+    private Random randomValuesSupplier = new Random();
+
+    public Ball getRandomBall() {
+        int randomInt = randomValuesSupplier.nextInt(LIMIT_OF_NUMBERS);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Rommelua, line 11: You have Random as a class field and ColorSupplier local in the method. What is the difference between these two objects? Why did you decide to do it this way?

Rommelua (changes requested):

src/main/java/core/basesyntax/Lottery.java
QbaSekowski: I made changes to Main() method and Lottery class.

QbaSekowski:
New changes made in Main() method and Lottery class.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int LIMIT_OF_NUMBERS = 101;
```
nowickimj, line 6: I think a wise approach is to separate business aspects (human logic) of code from technical ones. In this case - a business requirement is to assign a number up to 100 - so in my opinion, a constant should be assigned with value 100 instead of 101.
Then, a technical requirement is to provide a maximum value which will be used by `Random` object to select a number. In our case it will be `101` so when calling `nextInt()`, so you provide a constant + 1.
It is also important to provide a good, self-explanatory name for every constant and variable, in this case I think it would be better to name it "MAX_NUMBER".

nowickimj (changes requested):

src/main/java/core/basesyntax/Lottery.java
QbaSekowski: Thank you for your tips and explanation.

QbaSekowski:
Fixes after code review:
- fixed use of constant in Lottery class and constant's name for better understanding of the code

nowickimj (approved):

PR #1144, opened 2023-09-18
Develop

MinoDentori:
Add implementation of the "Lottery" task according to the task description.
Cleanup code check style audit.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,7 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall().toString());
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,10 +2,11 @@ package core.basesyntax;
 
 public class Application {
     private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         Lottery lottery = new Lottery();
         Ball[] balls = new Ball[NUMBER_OF_BALLS];
-        for(int i = 0; i < balls.length; i++) {
+        for (int i = 0; i < balls.length; i++) {
             balls[i] = lottery.getRandomBall();
             System.out.println(balls[i].toString());
         }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,11 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
     public static void main(String[] args) {
         Lottery lottery = new Lottery();
-        Ball firstRandomBall = lottery.getRandomBall();
-        Ball secondRandomBall = lottery.getRandomBall();
-        Ball thirdRandomBall = lottery.getRandomBall();
-        System.out.println(firstRandomBall + "\n" + secondRandomBall + "\n" + thirdRandomBall);
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for(int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
```
d1sam, line 11: System.out.println(someObject) calls that object's toString() function to convert the object to a string representation. That's why you don' need to call `.toString()`.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
+        Color[] colors = Color.values();
+        return colors[random.nextInt(colors.length)].name();
```
d1sam, line 9: NOT MISTAKE, but make a note: You realized method `getName()` in `enum Color` and you can use it for better colors' names.

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        String randomColor = colorSupplier.getRandomColor();
+        int randomNumber = getRandomNumber();
+        return new Ball(randomColor, randomNumber);
+    }
+
+    private int getRandomNumber() {
+        Random random = new Random();
```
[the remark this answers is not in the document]
MinoDentori: Fixed. If I get it in the right way, in this case, we put enum Color colors as a class member (constant) cause we potentially may use it in other methods inside ColorSuppliers.
But in general, when we build the architecture of a class, how do we decide, which way is better, class member, or local variable? For example, IDE often advises creating local variables, not fields, for such cases. Maybe you have some rule, from your experience, or some tips :)

src/main/java/core/basesyntax/ColorSupplier.java
MinoDentori: I guess I created it and used it in previous versions, but after that, I read the last point of "Common mistakes" and replaced it by just calling the name method.

src/main/java/core/basesyntax/Application.java
MinoDentori: Fixed. Did it too early. Screwed up with common mistakes sry.

MinoDentori:
Fixed comments, reread common mistakes.

src/main/java/core/basesyntax/Lottery.java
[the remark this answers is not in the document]
d1sam: Each call to your `getRandomNumber()` method will create a new `random` object, which is not a good practice. You can avoid this constant creation of this object by defining it at the class level. I'm waiting for this point to be fixed)

d1sam (changes requested):
Each call to your getRandomNumber() method will create a new random object, which is not a good practice. You can avoid this constant creation of this object by defining it at the class level. I'm waiting for this point to be fixed)

src/main/java/core/basesyntax/ColorSupplier.java
MinoDentori: Fixed. I thought that we needed a new generator to start a new sequence of pseudorandom numbers for each new Ball object, maybe I was confused with another case.

MinoDentori:
Fixed. I thought that we needed a new generator to start a new sequence of pseudorandom numbers for each new Ball object, maybe I was confused with another case.

src/main/java/core/basesyntax/Lottery.java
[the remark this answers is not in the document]
d1sam: Please declare `random` at the class level.

d1sam (changes requested):
Well done, but please declare `random` at the class level.

MinoDentori:
Fixed all comments. I hope.

d1sam (approved):
good job! You made a great work on your mistakes.

PR #1145, opened 2023-09-19
Solution of jv-lottery task

HovorukhaBohdan:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = {lottery.getRandomBall(), lottery.getRandomBall(), lottery.getRandomBall()};
+
+        for (Ball ball: balls) {
+            System.out.println(ball.toString());
```
mnyshenko, line 10: There is no need to call toString() explicitly inside System.out.println() method because it'll be called implicitly

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = {lottery.getRandomBall(), lottery.getRandomBall(), lottery.getRandomBall()};
```
mnyshenko, line 7: There is no need to create an array to satisfy the task condition. You can just use for loop to make to lottery.getRandomBall() 3 times, but write this statement only once inside a loop

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
mnyshenko, line 4: You have a special enum to represent different colors, please use it here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].name();
```
mnyshenko, line 8: 1) There is no need to return a string, instead return an enum instance here.
2) Also consider extracting new Random to the field on the class level, so you can reuse it and not create every time you call this method.
3) "Colors.values()" used twice here, consider extracting it to the field too

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(MAX_NUMBER));
```
mnyshenko, line 9: Please consider what should be moved to the class level here

mnyshenko (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Color[] color = Color.values();
```
mnyshenko, line 6: minor: be careful about naming, this variable holds all the colors, not only 1 color

mnyshenko (approved):

PR #1149, opened 2023-09-21
lottery task solving

BolsunovaNataliia:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
```
d1sam, line 5: Please don't add redundant empty lines!

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
...
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
```
d1sam, line 7: Don’t create several instances of the same class if you can use only one instance for your purposes.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        for (int i = 0; i < 3; i++) {
```
d1sam, line 6: Don't use magic number. What does it mean - 3?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        int number = new Random().nextInt(100);
```
d1sam, line 8: All magic numbers in your code should be constants. How do you think what max number will be returned by `random.nextInt(100);`?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+        Color color = Color.values()[index];
+
+        return String.valueOf(color);
```
d1sam, line 10: Don’t use toString() or String.valueOf()(it will call toString() under the hood) for getting String representation of enum constants. toString() is common for all enum constants.

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100;
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(color.getRandomColor(), random.nextInt(MAX_NUMBER));
```
bhdnchui, line 11: Please, pay attention that in that case the maximum number would be 99, but the task is to get 100.
I would recommend do not increase MAX_NUMBER, but add 1 to the `nextInt` method call result.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+    private Color color;
+    private int number;
+
+    public Ball(Color color, int number) {
+        this.setColor(color);
+        this.setNumber(number);
+    }
+
+    @Override
+    public String toString() {
```
bhdnchui, line 13: It would be better to place `toString` at the end of file. It looks more 'beautiful'.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int SIZE = 3;
```
bhdnchui, line 4: Please, let's look for a more suitable name.

bhdnchui (changes requested):
Great job!
But minor improvements are required.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+
+        return color;
```
d1sam, line 12: **Don’t create temporary variables if you can directly return value from the method**
Bad example:
```
        Color color = Color.values()[index];
        return color;
}
```
Improved example:
```
        return Color.values()[index];
}
```

d1sam (changes requested):
Good job, but please delete one redundant variable :)

PR #1150, opened 2023-09-21
Created a simple lottery.

ponomvrenko:
1. Create Ball class with colour & number private fields; constructor with colour and number parameters and overriding toString().
2. Added Enum Colours.
3. Added ColorSupplier class with getRandomColour() method.
4. Added Lottery class with getRandomBall() method.
5. And for testing, crete three balls and print info about them.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i < 4; i++) {
+            System.out.println("Ball №" + i);
+            System.out.println(lottery.getRandomBall());
+            System.out.println("----------");
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,18 @@
+package core.basesyntax;
+
+public class Ball {
+    private String colour;
+    private int number;
+
+    public Ball(String colour, int number) {
+        this.colour = colour;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Colour: " + this.colour
+                + "\n"
+                + "Number: " + this.number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colours.values().length);
+        return String.valueOf(Colours.values()[index]);
     }
 }
--- a/src/main/java/core/basesyntax/Colours.java
+++ b/src/main/java/core/basesyntax/Colours.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colours {
+    RED,
+    GREEN,
+    YELLOW,
+    ORANGE,
+    BLUE,
+    GREY,
+    BLACK
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        Random random = new Random();
+        return new Ball(supplier.getRandomColor(), random.nextInt(100));
+    }
+}


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i < 4; i++) {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colours.values().length);
```
d1sam, line 7: Don’t create several instances of the same class if you can use only one instance for your purposes. In each call of` getRandomColor() ` `new Random()` will be created.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        Random random = new Random();
+        return new Ball(supplier.getRandomColor(), random.nextInt(100));
```
d1sam, line 9: All magic numbers in your code should be constants. How do you think what max number will be returned by random.nextInt(100);?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier supplier = new ColorSupplier();
+        Random random = new Random();
```
d1sam, line 8: Think about which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/Colours.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colours {
```
d1sam, line 3: You have a method `getRandomColor()` (named without `u`). Make everywhere names without `u`.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Ball {
+    private String colour;
```
d1sam, line 4: You have a method `getRandomColor()` (named without `u`). Make everywhere names without `u`.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+    private int number;
+
+    public Ball(String colour, int number) {
+        this.colour = colour;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Colour: " + this.colour
```
d1sam, line 14: You have a method `getRandomColor()` (named without `u`). Make everywhere names without `u`.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return String.valueOf(Colours.values()[index]);
```

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
ponomvrenko: I got your point, a random integer between 0 (inclusive) and 100 (exclusive).. I'll remember it next time.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random random = new Random();
```
mnyshenko, line 6: This variable should not be static, each instance of ColorSupplier should have its own instance of Random class

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int NUMBER_BOUND = 101;
+    private final ColorSupplier supplier = new ColorSupplier();
```
mnyshenko, line 7: It's better to be more specific in naming, **supplier** is too abstract, colorSupplier would be a better choice

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public Ball getRandomBall() {
+        Random random = new Random();
```
mnyshenko, line 10: Still the issue with random. On each call of getRandomBall() you create a new instance of random, which is bad, lets put this instance on the class level, so we can reuse.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
+    private final int number;
+
+    public Ball(String colour, int number) {
```
mnyshenko, line 7: colour -> color. Here's a cool shortcut to replacing some string with another string everywhere in the project: https://www.jetbrains.com/help/idea/finding-and-replacing-text-in-project.html#replace_search_string_in_project

mnyshenko (changes requested):

d1sam (approved):
Well done!

PR #1152, opened 2023-09-21
created Ball class with color and number fields;

YevheniiVlasenko:
overrode method .toString;
created enum Color with colors;
implemented method getRandomColor;
implemented method getRandomBall;

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,13 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery newLottery = new Lottery();
+        Ball firstBall = newLottery.getRandomBall();
+        Ball secondBall = newLottery.getRandomBall();
+        Ball thirdBall = newLottery.getRandomBall();
+
+        System.out.println(firstBall);
+        System.out.println(secondBall);
+        System.out.println(thirdBall);
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+
+    }
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return this.color + " " + this.number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(100));
+    }
+}
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    RED,
+    ORANGE,
+    YELLOW,
+    GREEN,
+    BLUE,
+    INDIGO,
+    VIOLET,
+}


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery newLottery = new Lottery();
+        Ball firstBall = newLottery.getRandomBall();
+        Ball secondBall = newLottery.getRandomBall();
+        Ball thirdBall = newLottery.getRandomBall();
+
+        System.out.println(firstBall);
+        System.out.println(secondBall);
+        System.out.println(thirdBall);
```
d1sam, line 12: Use for loop for creating several objects of the same class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
d1sam, line 6: Don’t use static methods in your solution

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(100));
```
d1sam, line 7: Don’t use static methods in your solution
Static methods are in general a bad practice. Let’s better create an instance of a class which method you want to call. Think about which variables should be local in the method and which should be class-level.

d1sam, line 7: All magic numbers in your code should be constants. How do you think what max number will be returned by random.nextInt(100);?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+
```
d1sam, line 8: For which purpose did you create default constructor?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
```

d1sam (changes requested):

src/main/java/core/basesyntax/Application.java
YevheniiVlasenko: > Use for loop for creating several objects of the same class

are you sure we need a loop for 3 objects?

src/main/java/core/basesyntax/ColorSupplier.java
YevheniiVlasenko: > Think about which variables should be local in the method and which should be class-level.

Must avoid object creation in method (?)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_BAll_NUMBER = 101;
```
bhdnchui, line 6: Why it is `public`?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random randomIndex = new Random();
```
bhdnchui, line 6: Please, think about a more compatible name.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
bhdnchui, line 4: Please, store `Color` type value here instead of `String` type.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery newLottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
```
bhdnchui, line 7: Please, make a separate constant as it is a number from a task.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[randomIndex.nextInt(Color.values().length)].name();
```
bhdnchui, line 9: That line looks too long. Look for to a good practise [rule](https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html#:~:text=4.1%20Line%20Length,no%20more%20than%2070%20characters.).
Also, define `Color.values()` array as a class field. It would provide better performance.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random randomIndex = new Random();
+
     public String getRandomColor() {
```
bhdnchui, line 8: Make this method return `Color` type.

src/main/java/core/basesyntax/Lottery.java
bhdnchui: Also it is quite strange to store a variable with `101` value.
I think, it would be better to store 100 value and add 1 when you call `nextInt`,
```
MAX_BAll_NUMBER = 100;
...
random.nextInt(MAX_BAll_NUMBER + 1);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_BAll_NUMBER = 101;
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random numberCreator = new Random();
```
bhdnchui, line 8: Quite complicate name. It is not `numberCreator`. Think about more better name.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return this.color + " " + this.number;
```
bhdnchui, line 14: Please, provide more info here. At least add the name of the class.
You can also override toString with IntellijIdea. Try it.
`Ctrl + O / ⌃ O` Look for more shortcuts [here](https://resources.jetbrains.com/storage/products/intellij-idea/docs/IntelliJIDEA_ReferenceCard.pdf).

bhdnchui (changes requested):
Good job, but please, look at my comments.

d1sam (approved):
Well done!

PR #1154, opened 2023-09-21
Hw jv lottery

AdamMudrak:
Please pay attention to the fact that I haven't overridden toString() in Ball.java because I didn't get how to do this, why we need this and why in that exact class...

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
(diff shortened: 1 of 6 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
kozhukhovsky, line 7: move Random to the class field

kozhukhovsky (approved):

PR #1163, opened 2023-09-26
Study

DevIgork:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball firstLottery = Lottery.getRandomBall();
+        Ball secLottery = Lottery.getRandomBall();
+        Ball thrtLottery = Lottery.getRandomBall();
+        System.out.println(firstLottery);
+        System.out.println(secLottery);
+        System.out.println(thrtLottery);
```
d1sam, line 10: Use for loop for creating several objects of the same class

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,28 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+
```
d1sam, line 5: Please don't add redundant empty lines.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,23 @@
...
+
+public enum Colors {
+    Black("Black"),
+    White("White"),
+    Red("Red"),
+    Yellow("Yellow"),
+    Orange("Orange"),
+    Pink("Pink"),
+    Green("Green"),
+    Blue("Blue");
```
d1sam, line 11: Don’t forget how to name enums according to style guide (the same with naming of constant values)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
d1sam, line 6: Don’t use static methods in your solution
```suggestion
    public Ball getRandomBall() {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        Ball lotteryBall = new Ball();
+        lotteryBall.setColor(new ColorSupplier().getRandomColor());
+        lotteryBall.setNumber(new Random().nextInt(100));
```
d1sam, line 9: All magic numbers in your code should be constants. How do you think what max number will be returned by random.nextInt(100);?
Think about which variables should be local in the method and which should be class-level

d1sam (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
d1sam, line 5: All magic numbers in your code should be constants.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
...
+            System.out.println(new Lottery().getRandomBall());
```
d1sam, line 6: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Colors[] allColors;
+
+    public ColorSupplier() {
+        allColors = Colors.values();
+    }
```
d1sam, line 10: You can directly initialize you `allColors` class-field outside a constructor.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(allColors.length);
```
d1sam, line 13: Think about which variables should be local in the method and which should be class-level. new Random() object will be created on each getRandomBall() call. How can you avoid this?

d1sam (changes requested):
Good job, but there are some minor changes need to be done.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int LOTTERY_START_NUMBER = 3;
+    private static Lottery lottery = new Lottery();
```
mnyshenko, line 5: In this case, I suggest not putting the lottery on a class level, the main method will be called only 1 time, so it is better to put it there

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 101;
+    private Ball lotteryBall;
+
+    public Ball getRandomBall() {
+        lotteryBall = new Ball();
+        lotteryBall.setColor(new ColorSupplier().getRandomColor());
+        lotteryBall.setNumber(new Random().nextInt(MAX_BALL_NUMBER));
```
mnyshenko, line 12: Think about which variables should be local in the method and which should be class-level. You need to create a new Ball every time someone calls this method, so there is no point in storing a Ball on a class level because you cant reuse it in further invocations of this method. The opposite thing I can say about ColorSupplier and Random. You don't need to recreate them on each call of getRandomBall(), you can store them on the class level and reuse.

mnyshenko (changes requested):

mnyshenko (approved):

PR #1165, opened 2023-09-27
Hw2 oop solution

annton-shakur:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery ball1 = new Lottery();
+        Lottery ball2 = new Lottery();
+        Lottery ball3 = new Lottery();
+
+        System.out.println(ball1.getRandomBall());
+        System.out.println(ball2.getRandomBall());
+        System.out.println(ball3.getRandomBall());
```
d1sam, line 12: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

d1sam: Use for loop for creating several objects of the same class

Bad example:
```
public class Main {
    
      User firstUser = new User();
      User secondUser = new User();
      User thirdUser = new User();
      users[0] = firstUser;
      users[1] = secondUser;
      users[2] = thirdUser;
  }
}
```
Improved example:
```
public class Main {
    
          users[i] = new User();
      }
  }
}
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level. `new Random()` object will be created on each `getRandomColor() `call. How can you avoid this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        Ball ball = new Ball();
+        ColorSupplier color = new ColorSupplier();
```
d1sam, line 9: Also think about which variables should be local in the method and which should be class-level

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        Ball ball = new Ball();
+        ColorSupplier color = new ColorSupplier();
+        ball.setNumber(random.nextInt(100));
```
d1sam, line 10: All magic numbers in your code should be constants. How do you think what max number will be returned by` random.nextInt(100);`?

d1sam (changes requested):

annton-shakur:
Hi, thank you for the comments. I have implemented the mentioned changes. Though, I am not quite sure regarding the last comment of yours. Did you mean that I should increase the limit, so that number 100 could be also generated? 

"All magic numbers in your code should be constants." is kinda confusing. I would appreciate some explanation.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] balls = new Lottery[3];
+        for (int i = 0; i < 3; i++) {
```
mnyshenko, line 6: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier color = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        ball.setNumber(random.nextInt(101));
```
mnyshenko, line 11: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

mnyshenko (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
kshuryhin, line 6: I would recommend to make number of balls a constant value, as it will enhance readability.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    private static final int NUMBER_COUNT = 101;
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier color = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        ball.setNumber(random.nextInt(NUMBER_COUNT));
+        ball.setColor(color.getRandomColor());
+        return ball;
```
kshuryhin, line 14: I would recommend here to refactor this part of code. First of all, here we are using the same instance of `ball` per each `getRandomBall() ` method call. which potentially can cause some errors in the future. Also, generally it is not recommended to have partially-stated objects, means that we create our instance without meaningful data, and only at some point are setting `color` and `number`. So, I would recommend to create `new Ball(number, color)` constructor and return each time new ball.

kshuryhin (changes requested):
Good job, but still some points to consider

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < Ball.getBallNumber(); i++) {
```
kshuryhin, line 6: I would say constant of ball number should be better moved here, to the `Application` class, rather than to the `Ball`.

kshuryhin (approved):
Nice!

PR #1167, opened 2023-09-27
added a solution to the Lottery task

AgroFix:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 100;
+
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
```
d1sam, line 11: Think about which variables should be local in the method and which should be class-level. `new Random()` and `new ColorSupplier()` object will be created on each `getRandomBall()` call. How can you avoid this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 100;
```
d1sam, line 6: How do you think what max number will be returned by random.nextInt(MAX_VALUE), where MAX_VALUE = 100?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(MAX_VALUE));
+        return ball.toString();
```
d1sam, line 15: Use name() for getting String representation of enum constants
 ```
@Override
 public String() toString() {
    return "default";
 }
```

**Review by d1sam — CHANGES_REQUESTED** [review]

File: src/main/java/core/basesyntax/Colors.java [hunk 5]
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+enum Colors {
```
kshuryhin, line 3: We should consider using singular in enum names

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 102;
```
kshuryhin, line 6: For now, maximum possible random number is 101, but the task says it should be 100. How do you think we can fix that?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Ball ball = new Ball();
```
kshuryhin, line 9: I would recommend to make `ball` variable local-scoped. Also recommend to create constructor in `Ball` that will accept all fields and initialise them once object is created.

kshuryhin (changes requested):
Great job, just some minor points to consider

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class Lottery {
+    private static final int MAX_VALUE = 101;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor();
+        int number = random.nextInt(MAX_VALUE);
+        return new Ball(color, number);
```
bhdnchui, line 13: It would be better to rewrite that lines with the next one:
```
 return new Ball(colorSupplier.getRandomColor(), random.nextInt(MAX_VALUE));
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,15 @@
...
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String toString() {
```
bhdnchui, line 12: It would be great to add `@Override` annotation before overridden methods.

bhdnchui (approved):
Good job!

PR #1169, opened 2023-09-27
did a lottery code with random value

Vladlesn:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
+    Yellow,
+    Green,
+    Black,
+    White,
+    Orange,
+    Puprle;
```
d1sam, line 9: Don’t forget how to name constants according to style guide (the same with naming of enum values)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
```
d1sam, line 6: Don’t use static methods in your solution

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String clr = colorSupplier.getRandomColor();
+        int index = new Random().nextInt(100);
```
d1sam, line 9: All magic numbers in your code should be constants. How do you think what max number will be returned by `random.nextInt(100);`?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String clr = colorSupplier.getRandomColor();
```
d1sam, line 8: choose better name for this variable, e.g. `color`

src/main/java/core/basesyntax/Lottery.java
d1sam, line 9: Think about which variables should be local in the method and which should be class-level. `new Random()` and new `ColorSupplier()` object will be created on each `getRandomBall() `call. How can you avoid this?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        Colors color = Colors.values()[index];
+        return color.toString();
```
d1sam, line 9: Use name() for getting String representation of enum constants
```
 @Override
 public String() toString() {
    return "default";
 }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level. `new Random()` object will be created on each `getRandomColor(`) call. How can you avoid this?

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,6 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        System.out.println(Lottery.getRandomBall());
```
d1sam, line 5: To test the program, in` main()` method, create **three** balls using `getRandomBall() `method from `Lottery` class. And print information about them in the console.

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public Lottery() {
+        colorSupplier = new ColorSupplier();
+        random = new Random();
+    }
+
+    public Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor();
+        final int index = random.nextInt(100);
+        return new Ball(color, index);
```
bhdnchui, line 17: return new Ball(color, index);
```suggestion
        return new Ball(colorSupplier.getRandomColor(),  random.nextInt(100));
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier;
+    private Random random;
+
+    public Lottery() {
+        colorSupplier = new ColorSupplier();
+        random = new Random();
+    }
```
bhdnchui, line 12:
```suggestion
    private ColorSupplier colorSupplier = new ColorSupplier();
    private Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,21 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Lotterey lotterey = new Lotterey();
```
bhdnchui, line 7: You create new Lotterey on every getRandomColor method call. Try to avoid it.

bhdnchui (changes requested):
Good job! But a few comments need to be processed.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,22 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Lotterey lotterey = new Lotterey();
+
     public String getRandomColor() {
```
bhdnchui, line 8: Still `getRandomColor` will return the same color on every method call.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
bhdnchui, line 10: My bad that I didn't mark it before.
Make 100 as constant and think about which result you will receive.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,37 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+    }
```
bhdnchui, line 8: Do you need this constuctor?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,22 @@
...
-        return null;
+        return lotterey.getColor();
+    }
+
+    class Lotterey {
+        private Random random = new Random();
+
```
bhdnchui, line 14:
```suggestion
```

bhdnchui (changes requested):
Good job, bet few comments need to be processed.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private final int bound = 100;
```
bhdnchui, line 8: Make it constant.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(bound));
```
bhdnchui, line 11: How do you think what max number will be returned by `random.nextInt(100)`?
Is it compatible with task?

src/main/java/core/basesyntax/ColorSupplier.java
bhdnchui: ?

bhdnchui (changes requested):
Good job, but look at my comments.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,21 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
```
bhdnchui, line 6:
```suggestion
private final Lotterey lotterey = new Lotterey();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,21 @@
...
     public String getRandomColor() {
-        return null;
+        Lotterey lotterey = new Lotterey();
+        return lotterey.getColor();
```
bhdnchui, line 9:
```suggestion
        return lotterey.getColor();
```

bhdnchui (approved):
Good job.

PR #1170, opened 2023-09-27
implemented task according to README.md

vadymhrnk:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
```
d1sam, line 6: How do you think what max number will be returned by random.nextInt(MAX_BALL_NUMBER), where MAX_BALL_NUMBER = 100?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+    private ColorSupplier supplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(supplier.getRandomColor(), new Random().nextInt(MAX_BALL_NUMBER));
```
d1sam, line 10: Think about which variables should be local in the method and which should be class-level. `new Random()` object will be created on each `getRandomBall()` call. How can you avoid this?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
```
 @Override
 public String() toString() {
    return "default";
 }
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
d1sam, line 9: [Use for loop for creating several objects of the same class](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#use-for-loop-for-creating-several-objects-of-the-same-class)
Bad example:
```
public class Main {
    
      User firstUser = new User();
      User secondUser = new User();
      User thirdUser = new User();
      users[0] = firstUser;
      users[1] = secondUser;
      users[2] = thirdUser;
  }
}
```
Improved example:
```
public class Main {
    
          users[i] = new User();
      }
  }
}
```

d1sam (changes requested):

bhdnchui (approved):
Looks good to me.

PR #1171, opened 2023-09-27
Implemented Ball,Lottery classes, getRandomColor and getRandomBall me…

Andriy-Hrynevetskyi:
…thods + Color enum.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,11 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] lotteryBalls = new Ball[3];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < lotteryBalls.length; i++) {
+            lotteryBalls[i] = lottery.getRandomBall();
+            System.out.println(lotteryBalls[i]);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    BLUE,
+    GREEN,
+    RED,
+    YELLOW,
+    BLACK,
+    WHITE
+
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        Color randomColor = Color.values()[index];
+        return randomColor.toString();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        Random random = new Random();
+        Ball ball = new Ball();
+        ColorSupplier randomColor = new ColorSupplier();
+        ball.setNumber(random.nextInt(MAX_BALL_NUMBER));
+        ball.setColor(randomColor.getRandomColor());
+        return ball;
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color " + this.color + " Number " + this.number;
+    }
+}


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] lotteryBalls = new Ball[3];
```

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Color {
+    BLUE,
+    GREEN,
+    RED,
+    YELLOW,
+    BLACK,
+    WHITE
+
```
d1sam, line 10: Please do not add this empty redundant line.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Color randomColor = Color.values()[index];
+        return randomColor.toString();
```
d1sam, line 9: Don’t use toString() or String.valueOf()(it will call toString() under the hood) for getting String representation of enum constants. toString() is common for all enum constants. If you override this method like below:
 ```
@Override
 public String() toString() {
    return "default";
 }
```

File: src/main/java/core/basesyntax/Lottery.java [hunk 5]
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        Random random = new Random();
```
d1sam, line 9: Think about which variables should be local in the method and which should be class-level. new Random() object will be created on each getRandomBall() call. How can you avoid this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+        Ball ball = new Ball();
+        ColorSupplier randomColor = new ColorSupplier();
```
d1sam, line 11: Think about which variables should be local in the method and which should be class-level. new ColorSupplier() object will be created on each getRandomBall() call. How can you avoid this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
```
d1sam, line 6: Please be careful with bound of `random.nextInt(MAX_BALL_NUMBER)`. This method call returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and n (**exclusive**).

d1sam (changes requested):
good job, but there are some points need to be resolved.

Andriy-Hrynevetskyi:
I`ve made all improvements at the moment.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+    static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] lotteryBalls = new Ball[NUMBER_OF_BALLS];
```
bhdnchui, line 8: That array is redundant.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < lotteryBalls.length; i++) {
+            lotteryBalls[i] = lottery.getRandomBall();
+            System.out.println(lotteryBalls[i]);
```
bhdnchui, line 12: Rewrite that lines with the next one:
```
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 101;
+    private Random random = new Random();
+    private ColorSupplier randomColor = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setNumber(random.nextInt(MAX_BALL_NUMBER));
+        ball.setColor(randomColor.getRandomColor());
+        return ball;
```
bhdnchui, line 14: Let`s decrease the number of lines in your code. Replace those lines with the next one:
```
return new Ball(random.nextInt(MAX_BALL_NUMBER), randomColor.getRandomColor());
```
But you should add a such`public Ball(String color, int number){}` constructor to the ball class.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private int index = new Random().nextInt(Color.values().length);
+
     public String getRandomColor() {
-        return null;
+        Color randomColor = Color.values()[index];
+        return randomColor.name();
```
bhdnchui, line 10: Please, try to analize the result of the next code execution:
```
ColorSupplier supplier = new ColorSupplier();
for (int i = 0, i < 100; i++) {
    System.out.println(supplier. getRandomColor());
}
```
You should create a new random index for every `getRandomColor` method call.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private int index = new Random().nextInt(Color.values().length);
+
     public String getRandomColor() {
```
bhdnchui, line 8: Think about `Color` type return.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+    static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
bhdnchui, line 7: Remove that unnecessary comment.

bhdnchui (changes requested):
Good job, but please, pay attention to my comments.

d1sam (approved):
Well done!

PR #1172, opened 2023-09-27
Oop lottery task

Zakhar27:
Java OOP lesson, lottery task

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,10 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i <= 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public class Ball {
+    String color;
+    int number;
+
+    @Override
+    public String toString() {
+        return super.toString();
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,8 +1,8 @@
 package core.basesyntax;
 
 public class Ball {
-   public  String color;
-   public  int number;
+    private String color;
+    private int number;
 
     public Ball(Colors color, int number) {
         this.color = String.valueOf(color);
@@ -11,6 +11,6 @@ public class Ball {
 
     @Override
     public String toString() {
-        return "number - " + number +  "  " + "color - " + color;
+        return "number - " + number + "  " + "color - " + color;
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,11 +1,16 @@
 package core.basesyntax;
 
 public class Ball {
-    String color;
-    int number;
+   public  String color;
+   public  int number;
+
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(Colors color, int number) {
+        this.color = String.valueOf(color);
```
d1sam, line 8: **Use name() for getting String representation of enum constants**
 ```
@Override
 public String() toString() {
    return "default";
 }
```

File: src/main/java/core/basesyntax/ColorSupplier.java [hunk 3]
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level. new Random() object will be created on each getRandomBall() call. How can you avoid this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
d1sam, line 10: All magic numbers in your code should be constants.
Please see [this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names) to learn about constant fields and their naming requirements. How do you think what max number will be returned by `random.nextInt(100);`?

d1sam (changes requested):
Good job, but there are some points need to be resolved. Please look at my comments below.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "number - "
+                + number
+                + "  "
+                + "color - "
+                + color;
+    }
```
kshuryhin, line 19: I would recommend to place variables names on the same line with variables themselves, as it would be more readable. Like:
`"number - " + number`

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    static final int NUMBER = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        for (int i = 0; i <= NUMBER; i++) {
+            Ball ball = new Lottery().getRandomBall();
```
kshuryhin, line 9: Do we really need to create new `Lottery` instance during each iteration? Feel free to move `Lottery` creation out of the loop scope. 
Sorry for confusion, [this](https://github.com/mate-academy/jv-lottery/pull/1172#discussion_r1339603374) comment referred to magic numbers in the loop, which you replaced with `NUMBER` constant.

kshuryhin (changes requested):
Great job, just some minor stuff

src/main/java/core/basesyntax/Ball.java
Zakhar27: I overdid the rule with + every line)I changed that

kshuryhin (approved):
Cool!

PR #1173, opened 2023-09-27
Implemented tasks from README.md

DEz-1996:
Created Ball class;
Override method toString() in Ball class;
Created enum class BallsColors;
Implemented method getRandomColor() in class ColorSupplier; Created class Lottery;
Created loop for to call random balls in class Application.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        int numberOfBalls = 3;
+        for (int i = 0; i < numberOfBalls; i++) {
+            System.out.println(lottery.getRandomBall().toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        int numberOfBalls = 3;
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
...
+        for (int i = 0; i < numberOfBalls; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
d1sam, line 8: Please do not call `.toString()` inside `System.out.println()`, because it' called **under the hood** inside this method.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAXIMUM_POSSIBLE_BALL_NUMBER = 100;
```
d1sam, line 6: How do you think what max number will be returned by `random.nextInt(100);`?

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
DEz-1996: > How do you think what max number will be returned by `random.nextInt(100);`?

Thank you! Very observant advices!

kshuryhin (approved):
Good job!

PR #1174, opened 2023-09-27
Created all the needed classes. Realised their functionality.

EdenParadise:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,8 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
+        System.out.println(new Lottery().getRandomBall());
```
d1sam, line 8: [Don’t create several instances of the same class if you can use only one instance for your purposes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes).
**Also please use for loop for creating several objects of the same class**

Bad example:
```
public class Main {
    
      User firstUser = new User();
      User secondUser = new User();
      User thirdUser = new User();
      users[0] = firstUser;
      users[1] = secondUser;
      users[2] = thirdUser;
  }
}
```
Improved example:
```
public class Main {
    
          users[i] = new User();
      }
  }
}
```

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,14 @@
...
+    GREEN,
+    BLUE,
+    YELLOW,
+    ORANGE,
+    PURPLE,
+    PINK,
+    BLACK,
+    WHITE
+}
+
```
d1sam, line 14: Please do not add empty redundant lines.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
```
d1sam, line 7: Think about which variables should be local in the method and which should be class-level. new Random() object will be created on each getRandomBall() call. How can you avoid this?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        Color[] colors = Color.values();
+        Random random = new Random();
```
d1sam, line 8: Think about which variables should be local in the method and which should be class-level. new Random() object will be created on each getRandomColor() call. How can you avoid this?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+        int randomIndex = random.nextInt(colors.length);
+        Color randColor = colors[randomIndex];
+        return randColor.toString();
```
d1sam, line 11: Use name() for getting String representation of enum constants
 ```
@Override
 public String() toString() {
    return "default";
 }
```

File: src/main/java/core/basesyntax/ColorSupplier.java [hunk 6]
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
d1sam, line 6: Don’t use static methods in your solution

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
...
+        return new Ball(ColorSupplier.getRandomColor(), random.nextInt(100));
```
d1sam, line 8: All magic numbers in your code should be constants. How do you think what max number will be returned by random.nextInt(100);?

d1sam (changes requested):

src/main/java/core/basesyntax/Lottery.java
EdenParadise: I dont know. I have no clue. Absolutely

src/main/java/core/basesyntax/ColorSupplier.java
EdenParadise: I dont know. I have no clue. Absolutely

src/main/java/core/basesyntax/ColorSupplier.java
EdenParadise: Okay. I totally dont understand

src/main/java/core/basesyntax/ColorSupplier.java
EdenParadise: Why and how then if not static?

src/main/java/core/basesyntax/Lottery.java
EdenParadise: !!?!?!?!?!?!?!??!?!?!!?!??!?!?!?!

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0;i < 3;i++) {
+            System.out.println(new Lottery().getRandomBall());
```
d1sam, line 6: Don’t create several instances of the same class if you can use only one instance for your purposes
Bad example:
```
public class Main {
      String[] emails = {"myEmail@gmail.com", "not@Valid@.g.com"};
      for(String email : emails) {
          if (new EmailValidator().isValid(email)) {
              System.out.println("Email " + email + " is valid");
          }
      }
  }
```
}
Improved example:
```
public class Main {
      String[] emails = {"myEmail@gmail.com", "not@Valid@.g.com"};
      EmailValidator emailValidator = new EmailValidator();
      for(String email : emails) {
          if (emailValidator.isValid(email)) {
              System.out.println("Email " + email + " is valid");
          }
      }
  }
}

```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0;i < 3;i++) {
```
d1sam, line 5: All magic numbers in your code should be constants.

d1sam, line 5: Also please pay attention to formatting.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        final int magicNumber = random.nextInt(100);
```
d1sam, line 9: All magic numbers in your code should be constants.
Please see [this article](https://mate-academy.github.io/style-guides/java/java.html#s5.2.4-constant-names) to learn about constant fields and their naming requirements. It's about `100`

src/main/java/core/basesyntax/ColorSupplier.java
d1sam: This mistake is still exists.

d1sam (changes requested):
There are mistakes which not resolved. Please read [this](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery) article and try to avoid common mistakes.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    static final int BALLS_COUNT = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0;i < BALLS_COUNT;i++) {
```
bhdnchui, line 8: Format your code.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Color[] colors = Color.values();
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+
```
bhdnchui, line 10: Empty redundant line.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
+        Color randColor = colors[random.nextInt(colors.length)];
+        return randColor.toString();
```
bhdnchui, line 12:
```suggestion
        return colors[random.nextInt(colors.length)].name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        final int magicNumber = random.nextInt(BOUND);
```
bhdnchui, line 10: It is not `magicNumber`. Bad name :(

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    static final int BOUND = 100;
```
bhdnchui, line 6:
```suggestion
    private static final int BOUND = 100;
```

bhdnchui: Constants should be `private static final`.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        final int magicNumber = random.nextInt(BOUND);
+        return new Ball(new ColorSupplier().getRandomColor(), magicNumber);
```
bhdnchui, line 11: Make `ColorSupplier` instance class level field.

src/main/java/core/basesyntax/Lottery.java
bhdnchui: Remove that redundant temporary variable.

bhdnchui (changes requested):
Good job, but a few comments need to be processed.

bhdnchui (changes requested):
Nice! 
But a few comments need to be processed.

bhdnchui (approved):
Good job!

bhdnchui (changes requested):
Fix compilation error.

bhdnchui (approved):
Done.

PR #1175, opened 2023-09-28
did my homework

BohdanTrue:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private Ball ball = new Ball();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
bhdnchui, line 12: Please, follow this [rule](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants).

bhdnchui (changes requested):
Good job, but please, pay attention to my comments.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
...
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_USERS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (Ball ball: balls) {
+            System.out.println(ball.toString());
+        }
```
d1sam, line 15: Your array is redundant, because you can print info about 3 random balls without array and you can achieve this with only one cycle. Also please pay attention, that `ball.toString()` contains redundant call of `.toString()` method, because under-the-hood of `System.out.println()` we have a call of `.toString()` method of `ball`.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
```
d1sam, line 4: Why `USERS`? Maybe `NUMBER_OF_BALLS`

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
```
d1sam, line 31: Use name() for getting String representation of enum constants
 ```
@Override
 public String() toString() {
    return "default";
 }
```

File: src/main/java/core/basesyntax/Lottery.java [hunk 5]
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 100;
```
d1sam, line 6: How do you think what max number will be returned by random.nextInt(MAX_VALUE); with MAX_VALUE = 100?

d1sam (changes requested):
Good job, but some minor changes should be done.

d1sam (approved):
Well done!

PR #1178, opened 2023-09-28
created classes and implemented methods described in README.md; creat…

denysokhrymovych:
…ed 3 balls and printed information about them

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
```
mnyshenko, line 13: A few points here:
1) Please, pay attention to the task description, you don't need to store balls anywhere, you just need to create them and print information about them
2) No need to call toString() method inside System.out.println() because it'll be called implicitly

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
```
mnyshenko, line 7: 1) a new instance of random will be created on each invocation of getRandomColor() method, which is bad. It's better to store it at the class level.
2) Color.values() is used twice here, so it can be extracted on a class level too for further reuse of variables

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(101));
```
mnyshenko, line 9: 1. The same issue with random as inside ColorSupplier class
2. Please, pay attention to this common mistake: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#all-magic-numbers-in-your-code-should-be-constants

mnyshenko (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int UPPER_EXCLUSIVE_BOUND_FOR_BALL_NUMBER = 101;
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(),
+        random.nextInt(UPPER_EXCLUSIVE_BOUND_FOR_BALL_NUMBER));
```
bhdnchui, line 12: It would be great to place a `tab` here. In order to text on the 11 line is to the right of the text on the 12th.
```
return new Ball(colorSupplier.getRandomColor(),
        random.nextInt(UPPER_EXCLUSIVE_BOUND_FOR_BALL_NUMBER));
```

bhdnchui (approved):
Good job!

PR #1179, opened 2023-09-28
Made the task with Lottery

DPotapov93:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 public class Application {
+    private static final int countOfBalls = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[countOfBalls];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
+            System.out.println("№" + (i + 1) + "-" + balls[i]);
+        }
```
mnyshenko, line 11: 1) Please, pay attention to the task description. You don't need to store balls anywhere
2) Do you really need to create a new lottery each time you want to create a new ball?
3) Minor: I suggest just passing a newly generated ball to the println() method and letting to overridden toString() of the Ball class do all the work))

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int countOfBalls = 3;
```
mnyshenko, line 4: Don't forget about how we name constants: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#dont-forget-how-to-name-constants-according-to-style-guide-the-same-with-naming-of-enum-values

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
+        if (number < 0) {
+            System.out.println("number can`t be negative");
+        }
+        this.number = number;
```
mnyshenko, line 28: Are you sure we have such a limitation in this task?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[countOfBalls];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
+            System.out.println("№" + (i + 1) + "-" + balls[i]);
+        }
+
```
mnyshenko, line 12: https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery#please-dont-add-redundant-empty-lines-to-your-code

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int colorInt = new Random().nextInt(Colors.values().length);
+        return Colors.values()[colorInt].toString();
```
mnyshenko, line 8: 1) A new instance of random will be created on each invocation of getRandomColor() method, which is bad. It's better to store it at the class level
2) Colors.values() expression used twice here, it's better to store it on the class level too for further reuse

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
mnyshenko, line 8: 1) Again, point about creating Random. The same applies for colorSupplier
2) Don't forget about dealing with such numbers in the code (make them constants)
3) Please, pay attention to how nextInt() method works and what value you pass there

mnyshenko (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random colorRandom = new Random();
+    private int colorInt = colorRandom.nextInt(Colors.values().length);
```
d1sam, line 7: If you declare your `colorInt` variable at class-level it will be the same value in `getRandomColor()` method on each call. You should declare this variable at method-level (local).

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+
     public String getRandomColor() {
-        return null;
+        return Colors.values()[colorInt].toString();
```
d1sam, line 10: Use name() for getting String representation of enum constants
```
 @Override
 public String() toString() {
    return "default";
 }
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random random = new Random();
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private int randomInt = random.nextInt(100);
```
d1sam, line 8: If you declare your `randomInt` variable at class-level it will be the same value in `getRandomBall()` method on each call. You should declare this variable at method-level (local).
All magic numbers in your code should be constants. How do you think what max number will be returned by `random.nextInt(100);`?

d1sam (changes requested):
Good job, but there are some changes should be done.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colors {
```
mnyshenko, line 3: Enum's name should be written in singular Colors -> Color

mnyshenko (changes requested):

bhdnchui (approved):
Good job!

PR #1185, opened 2023-10-04
Homework

ddryzhov:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
...
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
okuzan, line 13: image

okuzan (changes requested):
Hey there! 😊 Before you submit your work for review, please go through the checklist under the task and make any necessary corrections. 📝✅ Your mentor will be happy to help once those points are sorted out.  Don't worry, it's all part of the learning process! 📚✨

src/main/java/core/basesyntax/Application.java
ddryzhov: fixed :-)

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public enum Colors {
+    RED,
```
okuzan, line 4:
```suggestion
public enum Color {
    RED,
```

image

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int i = new Random().nextInt(Colors.values().length);
+
+        return String.valueOf(Colors.values()[i]);
```
okuzan, line 9: image

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
```
```suggestion
            System.out.println(lottery.getRandomBall());
```

okuzan (changes requested):
Make Random (in both classes), ColorSupplier - class fields with appropriate access modifiers, not local variables which are recreated per each method invocation, Make 100 a class constant, enum name should be in singular, use .name() instead of String.valueOf. You don't need a ball array in the main - try to achieve the same without it :)

ddryzhov:
> Make Random (in both classes), ColorSupplier - class fields with appropriate access modifiers, not local variables which are recreated per each method invocation, Make 100 a class constant, enum name should be in singular, use .name() instead of String.valueOf. You don't need a ball array in the main - try to achieve the same without it :)

done

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int oneHundred = 100;
```
```suggestion
    private static final int QUANTITY_LIMIT = 100;
```
don't name constants with their value - use constant name to tell people who will read your code about the role it serves in the application,

```suggestion
    private static final int oneHundred = 100;
    private ColorSupplier = new ColorSupplier();
    private Random = new Random();
```
make these objects belong to class instance - they will be created only once for each instance, per your current implementation they are recreated each time the method is called

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        return Color.values()[new Random().nextInt(Color.values().length)];
     }
```
okuzan, line 8: extract random to a class field here too, same as in lottery class

okuzan (changes requested):

ddryzhov:
fixed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Ball ball = new Ball();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(QUANTITY_LIMIT));
+
+        return ball;
+    }
```
Ivan95kos, line 16: It is better to create a `new Ball `each time
```suggestion
    private final Random random = new Random();

    public Ball getRandomBall() {
        Ball ball = new Ball();
        ball.setColor(colorSupplier.getRandomColor());
        ball.setNumber(random.nextInt(QUANTITY_LIMIT));

        return ball;
    }

```

Ivan95kos (approved):
Great job, with minor comment

PR #1197, opened 2023-10-10
add class Ball with fields and constructor, add Enum Color, add class…

EugeneKorniuk:
… Lottery

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,17 @@
+package core.basesyntax;
+
+public class Ball {
+    String color;
+    int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color " + color
+                + ", number " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
+        int index = new Random().nextInt(Color.values().length);
+        Color color = Color.values()[index];
+        return color.toString();
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color " + color
+                + ", number " + number;
```
karma-o, line 31: Let's use the following format, which is default if we use IDE `toString()` generation feature
```suggestion
        return "Ball{"
                + "color=" + color
                + ", number=" + number
                + "}";
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
+        int index = new Random().nextInt(Color.values().length);
```
karma-o, line 8: Make random an instance level variable to not create a new one each time
`Color.values()` call result can be saved to a variable to avoid multiple calls

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
```
karma-o, line 8: same for these variables

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100));
```
karma-o, line 9: check out the documentation of the `nextInt()` method. The bound parameter there is exclusive, which means max number that we can get here is 99.
Also, make this number a constant

karma-o (changes requested):

PR #1200, opened 2023-10-11
Realized simple lottery - added Lottery class with getRandomBall() me…

SSosedova:
…thod.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,9 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i <= 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 7: make Random a class field and you won't unnecessarily recreate it per each method execution :)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
```
okuzan, line 8: same for these two

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(101));
```
okuzan, line 9: make 101 a class constant

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i <= 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
olekskov, line 8: toString() method is automatically called when we pass objects to System.out.println
```suggestion
            System.out.println(lottery.getRandomBall());
```

olekskov (approved):
Good job

PR #1206, opened 2023-10-14
Done task Lottery. Created classes Ball, Lottery, fill class ColorSup…

AntonBabychP1T:
…plier, added ENUM Colors and in the Application 3 outputs to console with info about random balls.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return (Colors.values()[new Random().nextInt(
+                Colors.values().length)]).toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
+    RED,
+    BLUE,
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+
```
Sanderfuck, line 9: Repeating code, maybe better use loop?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return (Colors.values()[new Random().nextInt(
+                Colors.values().length)]).toString();
```
Sanderfuck, line 8: Every invoke will make new object of Random, you need only one, what about field? In enum you can use name instead of toString

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int RANDOM_BOUND = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return (new Ball(colorSupplier.getRandomColor(), new Random().nextInt(RANDOM_BOUND)));
+    }
```
Sanderfuck, line 11: The same thing, you need only one instance of ColorSupplier and Random

Sanderfuck (changes requested):
Correct what in comments

src/main/java/core/basesyntax/Application.java
AntonBabychP1T: Thank you for feedback, i commit and push new changes, but i think something wrong with my git, because new commit did not pass automatic verification. Please, check it in commits of this project. And i would be glad if you could explain if I did everything right when interacting with git

Rommelua: All questions should be discussed in chat, not in the PR

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

Rommelua (approved):
Good job, one comment

PR #1214, opened 2023-10-24
Completed the task of creating random balls

AnnaSaienko:


src/main/java/core/basesyntax/FigureSupplier.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class FigureSupplier {
```
Rommelua, line 5: FigureSupplier? What is it? We are dealing with Lottery and Ball here

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private enum Colors {
```
Rommelua, line 6: Use name **Color**
Make it public enum in a separate class

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public class Lottery {
+    private final FigureSupplier figureSupplier = new FigureSupplier();
```
Rommelua, line 4: add Random field and delete FigureSupplier at all

Rommelua (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
+    private static final Lottery lottery = new Lottery();
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_USERS];
+        for (int i = 0; i < NUMBER_OF_USERS; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i].toString());
```
Sanderfuck, line 11: What will be change if you don`t use method .toString()?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(Color.values().length)].name();
```
Sanderfuck, line 7: Every invoke of method create a new object of random, but we need only one like in Lottery class

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int MAX_NUMBER = 101;
```
Sanderfuck, line 6: Why you write 101 if we need 100?

Sanderfuck (changes requested):

src/main/java/core/basesyntax/Lottery.java
AnnaSaienko: As I understood, according to the task, the maximum possible number on the ball should be 100, so in the .nextInt() method, I used the number 101 as the limit that is excluded. But maybe I just misunderstood this part of the task.

PR #1217, opened 2023-10-25
adding basic functionality and classes

bohdan2704:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
...
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int numberOfBalls = 3;
+        final int maxNumberOnTheBall = 100;
+        Ball[] balls = new Ball[numberOfBalls];
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall(maxNumberOnTheBall);
+            System.out.printf("%d. %s\n", i + 1, balls[i].toString());
```
Sanderfuck, line 15: Do you need toString method here?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,20 @@
...
+
+    public ColorSupplier() {
+        // SHOULD I CREATE RANDOM OBJECT AS FIELD FOR ALL CLASS,
+        // OR IT IS BETTER TO LEAVE IT AS VARIABLE IN METHOD getRandomColor ?
+        this.random = new Random();
+    }
+
     public String getRandomColor() {
-        return null;
+        int r = random.nextInt(Color.values().length);
```
Sanderfuck, line 16: Bad naming "r"

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,21 @@
...
+    private final Random random;
+
+    public Lottery() {
+        // SHOULD I MAKE RANDOM AS CLASS FIELD OR JUST AS SIMPLE LOCAL FUNCTION VARIABLE ?
+        this.random = new Random();
+    }
+
+    public Ball getRandomBall(int maxNum) {
+        // SHOULD maxNum be included in the range or not ?
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Sanderfuck, line 16: Every invoke of this method create new object ColorSupplier

Sanderfuck (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,22 @@
...
+    private final Random random;
+    private final ColorSupplier colorSupplier;
+
+    public Lottery() {
+        // SHOULD I MAKE RANDOM AS CLASS FIELD OR JUST AS SIMPLE LOCAL FUNCTION VARIABLE ?
+        this.random = new Random();
+        colorSupplier = new ColorSupplier();
+    }
+
+    public Ball getRandomBall(int maxNum) {
```
kerrrusha, line 16: you dont need any parameters for this method
```suggestion
    public Ball getRandomBall() {
```
image

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class Lottery {
+    private final Random random;
+    private final ColorSupplier colorSupplier;
+
+    public Lottery() {
+        // SHOULD I MAKE RANDOM AS CLASS FIELD OR JUST AS SIMPLE LOCAL FUNCTION VARIABLE ?
+        this.random = new Random();
+        colorSupplier = new ColorSupplier();
+    }
```
kerrrusha, line 14: we should make this fields as class fields because there's no purpose in creating them each time when getRandomBall() method is being invoked. we need to create them just once 
```suggestion
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public Ball getRandomBall(int maxNum) {
+        // SHOULD maxNum be included in the range or not ?
```
kerrrusha, line 17: according to the task:
"the maximum possible number should be 100"
so yes, it should be included.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,20 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random;
+
+    public ColorSupplier() {
+        // SHOULD I CREATE RANDOM OBJECT AS FIELD FOR ALL CLASS,
+        // OR IT IS BETTER TO LEAVE IT AS VARIABLE IN METHOD getRandomColor ?
+        this.random = new Random();
+    }
```
kerrrusha, line 13: we should make this fields as class fields because there's no purpose in creating them each time when getRandomColor() method is being invoked. we need to create them just once
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,20 @@
...
+
     public String getRandomColor() {
-        return null;
+        int randomColorCode = random.nextInt(Color.values().length);
```
kerrrusha, line 16: its not the code but index

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
 package core.basesyntax;
 
+import core.basesyntax.objects.Ball;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int numberOfBalls = 3;
+        final int maxNumberOnTheBall = 100;
```
kerrrusha, line 9:
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
 package core.basesyntax;
 
+import core.basesyntax.objects.Ball;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int numberOfBalls = 3;
```
kerrrusha, line 8: make it `private static final` constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
...
+        final int maxNumberOnTheBall = 100;
+        Ball[] balls = new Ball[numberOfBalls];
```
kerrrusha, line 10: it's not critical, but we can avoid creating array and just create objects and print them

kerrrusha (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,23 @@
+package core.basesyntax;
+
+import core.basesyntax.objects.Ball;
+import java.util.Random;
+
+public class Lottery {
+    private final Random random;
+    private final int maxNumberOnTheBall;
```
Sanderfuck, line 8: It can be constant in this case

Sanderfuck (changes requested):

src/main/java/core/basesyntax/Lottery.java
bohdan2704: What can be constant in this case ? Two fields are already final. Class is not used in inheritance, so there is no need for final class.

bohdan2704: Not understand what 'It can be constant in this case'. Fields are already final, should they be public static ?

Sanderfuck (approved):

PR #1224, opened 2023-11-01
implemented Lottery logic program

kriznychenko:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lotteryArray = new Lottery[3];
```
kozhukhovsky, line 5: No need for multiple Lottery objects. One is enough.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+        String randomColor = color.name();
+        return randomColor;
```
kozhukhovsky, line 12:
```suggestion
        return color.name();
```

kozhukhovsky (changes requested):
Good job. Need rework main method.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,7 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
```
Ivan95kos, line 5: one minor mistake:
Знімок екрана 2023-11-01 о 19 40 00

Ivan95kos (approved):
Great job, with a minor comment

PR #1231, opened 2023-11-05
lottery solution

mykolatsupryk:


kerrrusha (changes requested):

src/main/java/core/basesyntax/model/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax.model;
+
+import core.basesyntax.ColorSupplier;
+import core.basesyntax.enums.Color;
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        return new Ball(new Random().nextInt(101),
```
kerrrusha, line 10: magic number, need constant for that
```suggestion
        return new Ball(new Random().nextInt(MAX_NUMBER),
```

kerrrusha (approved):
good job

PR #1236, opened 2023-11-08
created a simple lottery

YaroslavRadevychVynnytskyi:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball.toString());
```
Rommelua, line 8:
```suggestion
            System.out.println(ball);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
Rommelua, line 6: image

Rommelua (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private int index = new Random().nextInt(Colors.values().length);
```
Rommelua, line 6: With this approach, ColorSupplier will return the same color each time
Random have to be class field

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Rommelua, line 8: see common mistakes

Rommelua (changes requested):

src/main/java/core/basesyntax/Lottery.java
YaroslavRadevychVynnytskyi: Is the problem here in the empty line 6 or in the return statement of the getRandomBall method?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+    private int index = random.nextInt(Color.values().length);
```
sokimaaa, line 7: Did you test your solution? Did the getRandomColor truly return random color?

This field would be the same for the same object, so make it a local variable 
```suggestion
    private int index = random.nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
sokimaaa: Color Supplier should be instantiated once - think about making it class level field. 

The same for Random. 

100 is a magic number, use constant to avoid direct using of magic numbers.

sokimaaa (changes requested):

src/main/java/core/basesyntax/Lottery.java
YaroslavRadevychVynnytskyi: Thank you. I think I got it

Rommelua (approved):
well done

PR #1250, opened 2023-11-18
task completed

YaroslavBerezhniak:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
olekskov, line 14: Please use cycle to perform repetitive actions

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import core.basesyntax.classes.Color;
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 8: Please extract random to class variable in order to avoid new object creation each time method is called

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+        Color color = Color.values()[index];
+        return color.toString();
```
olekskov, line 10: we should use .name() method to get String representation of enum

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import core.basesyntax.classes.Color;
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
olekskov, line 7: Please do not use static methods in your solution.
Static methods are generally much harder to maintain and test so should be used only when static is required

src/main/java/core/basesyntax/classes/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax.classes;
+
+import core.basesyntax.ColorSupplier;
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        String randomColor = ColorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(100) + 1;
```
olekskov, line 9: Exctract random and color supplier to be class variables

olekskov (changes requested):

PR #1259, opened 2023-11-27
all classes were created and methods were implemented

Artem-Kushniruk:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball firstBall = new Lottery().getRandomBall();
+        Ball secondBall = new Lottery().getRandomBall();
+        Ball thirdBall = new Lottery().getRandomBall();
+
+        System.out.println(firstBall.toString());
+        System.out.println(secondBall.toString());
+        System.out.println(thirdBall.toString());
```
kerrrusha, line 11: DRY - dont repeat yourself. code duplicating is a bad practice and can cause bugs. use loop instead

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+        return number;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
```
kerrrusha, line 26: it's better not to implement what not required
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
```
kerrrusha, line 5: `Random` should be `private final` class-level field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+    private final int index = new Random().nextInt(Colors.values().length);
```
kerrrusha, line 6: it's not random really. now, you are creating index once and use that index all the time, so `getRandomColor` would return single color, not random. consider creating index in `getRandomColor` method

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+
     public String getRandomColor() {
-        return null;
+        return Colors.values()[index].toString();
```
kerrrusha, line 9: read `checklist.md` about using `.toString()` or `.name()` with enums, and make corresponding fix

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final String color = new ColorSupplier().getRandomColor();
+    private final int number = new Random().nextInt(100);
```
kerrrusha, line 7: `ColorSupplier` and `Random` should be `private final` class-level fields

kerrrusha (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colors {
```
sokimaaa, line 3: Enum should never be in plural form
```suggestion
public enum Color {
```

sokimaaa (changes requested):

PR #1269, opened 2023-12-03
added_hw-jv-lottery-solution

SkillfulDev:
hw-jv-lottery-solution

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,10 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    public static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        for (int i = 0; i < 3; i++) {
-            System.out.println(new Lottery().getRandomBall());
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            System.out.println(lottery.getRandomBall());
         }
     }
 }
-
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -5,10 +5,10 @@ import java.util.Random;
 public class Lottery {
     private static final Random random = new Random();
     private static final ColorSupplier colorSupplier = new ColorSupplier();
-    private static final int MAX_POSSIBLE_NUMBER = 101;
+    private static final int MAX_NUMBER_EXCLUSIVE = 101;
 
     public Ball getRandomBall() {
-        int randomNumber = random.nextInt(MAX_POSSIBLE_NUMBER);
+        int randomNumber = random.nextInt(MAX_NUMBER_EXCLUSIVE);
         return new Ball(colorSupplier.getRandomColor(), randomNumber);
     }
 }


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] ballsArray = {new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()};
```
karma-o, line 7: Do we need 3 lotteries to get 3 balls? :)
We also can create them in the same loop that we have to print them

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,36 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+
```
karma-o, line 5: this line is redundant
```suggestion
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,36 @@
...
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball color is "
+                + color
+                + " and number is "
+                + number;
+    }
```
karma-o, line 19: `toString()` usually goes at the end of the class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
+        int randomColor = random.nextInt(Color.values().length);
+        return Color.values()[randomColor].name();
```
karma-o, line 9: random can be a class field to reuse it,
`Color.values()` can be a method variable to avoid multiple calls

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int randomNumber = random.nextInt(100);
+        ColorSupplier colorSupplier = new ColorSupplier();
```
karma-o, line 9: `random` and `colorSupplier` should be a class field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int randomNumber = random.nextInt(100);
```
karma-o, line 8: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99
And make it a constant

karma-o (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
+        }
     }
 }
+
```
karma-o, line 10: redundant empty line, should be only 1 at the end of the file
Hint: check in the IDE because github shows it a bit different 😉

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 5: 3 could also be a constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
...
+            System.out.println(new Lottery().getRandomBall());
```
karma-o, line 6: we are still creating a new lottery on each iteration, we should initialize it outside of for loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+    private static final int MAX_POSSIBLE_NUMBER = 101;
```
karma-o, line 8: naming is a bit confusing now because it says that the max possible number is 101, when in fact max number is 100, we should think about how we can make it more clear and readable

karma-o (changes requested):

karma-o (approved):
Nice! Also, next time check the common mistakes in the task description 😉

PR #1271, opened 2023-12-03
Created Ball class with fields color and number.

TarasYashchuk:
Override toString method.
Created Enum with different colors
Implemented getRandomColor method which returned random color based on enum values. Created Lottery class with getRandomBall method, which creates ball with a random color and a random number. In main class created three instances of ball using getRandomBall method and printed information about them

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,19 @@
 package core.basesyntax;
 
 public class Application {
+
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        Lottery[] lotteries = new Lottery[NUMBER_OF_BALLS];
+
+        for (int i = 0; i < lotteries.length; ++i) {
+            lotteries[i] = new Lottery();
+        }
+
+        for (Lottery lottery : lotteries) {
+            System.out.println(lottery.getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
+        Color color = Color.values()[new Random().nextInt(Color.values().length)];
+
+        return color.name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private static final int MAX_NUM = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+
+        String color = colorSupplier.getRandomColor();
+
+        return new Ball(color, new Random().nextInt(MAX_NUM));
+    }
+}
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,30 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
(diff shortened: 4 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private static final int MAX_NUM = 100;
```
Rommelua, line 7: Please don't add redundant empty lines to your code.
We don't need them after class declaration or method signature.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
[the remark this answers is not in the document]
TarasYashchuk: So, if some number (for example MAX_NUM ) is used only in some one method, and this value does not need to be shared among other methods of the class, is it better to make it local to the method?

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
Sanderfuck, line 9: image

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Color color = Color.values()[new Random().nextInt(Color.values().length)];
```
Sanderfuck, line 6: Init this logic inside method

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        int maxValue = 100;
```
Sanderfuck, line 8: You need to save it in constant

src/main/java/core/basesyntax/Lottery.java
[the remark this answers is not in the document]
Sanderfuck: image

Sanderfuck (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+
+        Color color = Color.values()[new Random().nextInt(Color.values().length)];
```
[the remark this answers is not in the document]
sokimaaa: not fixed

src/main/java/core/basesyntax/Lottery.java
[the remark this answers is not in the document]
sokimaaa: not fixed

sokimaaa (changes requested):
not fixed prev comments

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+    private Color color;
```
Sanderfuck, line 7: You don`t need Color field, you are not create object of this

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+
     public String getRandomColor() {
-        return null;
+        color = Color.values()[random.nextInt(Color.values().length)];
+        return color.name();
```
Sanderfuck, line 11:
```suggestion
        int randomValue = random.nextInt(Color.values().length);
        return Color.values()[randomValue].name();
```

Sanderfuck (changes requested):

Sanderfuck (approved):

PR #1273, opened 2023-12-05
jv-lottery solution

ValentynSharshon:
1. created Color enum with available ball colors;
2. created Ball class with color and number fields, constructor with both parameters and toString() method;
3. implemented getRandomColor() method in ColorSupplier class, witch return random color;
4. created NumberSupplier class with getRandomNumber() method, witch return random number in a given range;
5. created Lottery class with 'colorSupplier' and 'numberSupplier' fields, getRandomBall() method, witch return a Ball with random color and number;
6. tested the work of the programme in the main method of Application class.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,21 @@
 package core.basesyntax;
 
+import static java.lang.System.out;
+
 public class Application {
+
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (Ball ball : balls) {
+            out.println(ball);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public enum Color {
+
+    RED,
+    GREEN,
+    YELLOW,
+    ORANGE,
+    BLUE,
+    WHITE,
+    BLACK
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int randomColorIndex = random.nextInt(Color.values().length);
+        return Color.values()[randomColorIndex].name();
     }
 }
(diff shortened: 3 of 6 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,21 @@
...
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (Ball ball : balls) {
+            out.println(ball);
+        }
```
karma-o, line 19: we can remove the array and leave just 1 loop where we print the result of `getRandomBall()` right away, but I guess this approach is Ok too :)

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Color {
+
```
karma-o, line 4:
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int randomColorIndex = random.nextInt(Color.values().length);
+        return Color.values()[randomColorIndex].name();
```
karma-o, line 11: `Color.values()` call result can be saved to a variable to avoid multiple calls

src/main/java/core/basesyntax/NumberSupplier.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class NumberSupplier {
```
karma-o, line 5: I'd say that the `Random` class is a number supplier by itself, so we can just use it straight without this class.
Plus this was not part of the task so let's refactor your solution just a bit, although I like your effort to do a clear separation of responsibility :+1:

src/main/java/core/basesyntax/NumberSupplier.java
```diff
@@ -0,0 +1,13 @@
...
+    private static final int NUMBER_BOUND = 100;
```
karma-o, line 6: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99

karma-o (changes requested):

src/main/java/core/basesyntax/NumberSupplier.java
ValentynSharshon: Sorry, my mistake, fixed it.

src/main/java/core/basesyntax/ColorSupplier.java
ValentynSharshon: Made Color.values() a separate variable.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_BALLS];
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
```
karma-o, line 12: you see, now we do not really need that array

karma-o (approved):
Good job!

PR #1274, opened 2023-12-05
task completed with error correction

YaroslavBerezhniak:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+        Lottery lottery = new Lottery(colorSupplier, random);
```
Sanderfuck, line 9: If you send Random object in Lottery constructor, why you don`t do it for ColorSupplier?

Sanderfuck (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
+
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
```
okuzan, line 13: no need to create a local variable that is immediately recycled
```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

okuzan (changes requested):

okuzan (approved):
Good job, keep it up! 💪

PR #1275, opened 2023-12-07
realized 3 classes, implemented getRandomColor method

Oleksii-Naumenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
kozhukhovsky, line 7: move Random to the class field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Color color = Color.values()[index];
+        return color.toString();
```
kozhukhovsky, line 9:
```suggestion
        return color.name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int ballNumber = new Random().nextInt(100);
```
kozhukhovsky, line 7: move Random and ColorSupplier to the class field
move 100 to the constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < balls.length; i++) {
```
okuzan, line 8: rewrite without an array and using one for loop

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: remove filler comments before committing your solution

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
okuzan: no need to store it in the array, use right away -  `System.out.println(lottery.getRandomBall());`

okuzan (changes requested):

okuzan (approved):

PR #1277, opened 2023-12-10
Fixed problems with commit and push all project files

vfilimo:


vfilimo:
Fixed problems with commit and push all project files

PR #1278, opened 2023-12-10
taskReadyToCheck

shumYevh:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomColor() {
+
+        int ballNumber = new Random().nextInt(1, 100);
```
okuzan, line 11:
```suggestion
        int ballNumber = random.nextInt(1, 101);
```
99 would be the biggest generated

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
```
```suggestion
```
no empty line after class declaration needed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomColor() {
+
```
okuzan, line 10: no empty line after method declaration needed, fix everywhere

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
```
okuzan, line 7: make Random a class field too, like color supplier

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,53 @@
+package core.basesyntax;
+
+public class Ball {
+
```
okuzan, line 4:
```suggestion

```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,53 @@
...
+    private boolean isColor(String color) {
+
+        boolean checkColor = false;
+        for (int i = 0; i < Color.values().length; i++) {
+            if (color.equals(Color.values()[i].toString())) {
+                checkColor = true;
+            }
+        }
+        return checkColor;
+    }
```
okuzan, line 46: ball class shouldn't be bothered with logic of color supplier (like knowing limits for number 0 to 100), This class should only make it possible to create a ball instance given some name, and some color
```suggestion
    public Ball(int number, String color) {
         this.number = number;
         this.color = color;
     }   

```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,7 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomColor().toString());
```
okuzan, line 7: toString will be called behind the scenes by default, create a for loop - we should print more than one ball

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 5; i++) {
+            System.out.println(lottery.getRandomColor() + "\n");
```
```suggestion
            System.out.println(lottery.getRandomColor());
```
better use System.lineSeparator() than \n, however we don't need anything here - println prints a line + newline symbol

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomColor() {
+        int ballNumber = random.nextInt(1, 101);
```
okuzan, line 10: let's make 101 a MAX_BOUND constant, don't forget we use CONSTANT_CASE to name them

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 7: let's make Random a private class field, it will be more efficient than creating a separate instance for each color randomization

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,25 @@
...
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball number is : " + number + "\nBall color is : " + color;
+    }
+}
+
```
okuzan, line 25:
```suggestion

```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: remove comment

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
okuzan, line 10: sorry, forgot to point out [this](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants)

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
shumYevh: understood

PR #1291, opened 2023-12-17
created a ball class and implemented method getRandomBall

Bohadan-Markatov:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        for (int i = 1; i <= 3; i++) {
```
karma-o, line 6: `3` can be a constant here

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private int number;
+
+    public Ball(Color color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String toString() {
+        return "Ball color = " + color + ", "
+                + "Ball number = " + number;
```
karma-o, line 14: I'd suggest using the default format of `toString()` to avoid mentioning `Ball` each time
```suggestion
        return "Ball{" 
                + "color=" + color
                + ", number=" + number
                + "}";

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
karma-o, line 7: make Random an instance-level variable of `ColorSupplier`

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Color.values()[index];
```
karma-o, line 8: `Color.values()` can be saved to a var to avoid multiple calls

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        Color randomColor = ColorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(100);
```
karma-o, line 8: do not create a new `ColorSupplier` and `Random` each time

karma-o, line 8: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99.
And make it a constant

karma-o (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static Random random = new Random();
+
+    public static Color getRandomColor() {
```
karma-o, line 8: We should also avoid static methods and variables, the same for other classes + check common mistakes for more details.
And I wanted to mention that it is not the guided way to change the return type and method signature, but usually, it is a proper way to store Data in enum type instead of string in `Ball` class, so we can leave it as it is

karma-o (approved):
Approved, but left a comment and please check common mistakes before sending your solution to review, next time it won't be reviewed until those mistakes are fixed

karma-o (approved):

PR #1297, opened 2023-12-23
[jv-lottery]: home-work solution

soqiable:
image

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+    }
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Random ball with color - " + color + " and number: " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[index]);
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    public Ball getRandomBall() {
+        String randomColor = new ColorSupplier().getRandomColor();
+        int randomNumber = new Random().nextInt(100);
```
vovasalyha, line 8: Try to avoid magic numbers.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return String.valueOf(Colors.values()[index]);
```
vovasalyha, line 8: You can use `name()` method of an enum instead of `String.valueOf()`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends Ball {
+    public Ball getRandomBall() {
+        String randomColor = new ColorSupplier().getRandomColor();
```
vovasalyha, line 7: ColorSupplier and Random can be fields instead of method variables.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
```
vovasalyha, line 7: Do you need this constructor somewhere?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
```
vovasalyha, line 4: If fields are not intended to be changed it's better to mark them as `final`

vovasalyha (changes requested):
Looks good, but see a couple of comments.

vovasalyha (approved):

PR #1300, opened 2023-12-26
HW lottery solution

YuliiaNisha:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
```
okuzan, line 8: make them two private class fields

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
```
```suggestion
    public Ball getRandomBall() {
```
This makes codes more flexible. If we need to get a string representation of a ball, we can get it at any time, but if we want a ball itself? We wouldn't be able to get it.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(100)).toString();
```
okuzan, line 9: make 100 a class constant, use CONSTANT CASE
Also, let's change it to 101 (passed value is excluded) - to make 100 a maximum possible number

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 7: random - to a class field

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
okuzan, line 8: use a `for` loop

okuzan (changes requested):

Sanderfuck (approved):

PR #1301, opened 2023-12-29
ball-constructor-declaration-fixed

MolchanovAlexander:


MolchanovAlexander:
Is anybody here?

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball b1 = new Lottery().getRandomBall();
+        Ball b2 = new Lottery().getRandomBall();
+        Ball b3 = new Lottery().getRandomBall();
+        System.out.println(b1.toString());
+        System.out.println(b2.toString());
+        System.out.println(b3.toString());
```
karma-o, line 11: Do we need 3 lotteries to get 3 balls? :)
Also let's use loops for this logic

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private int number;
+
+    Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return this.color + " " + this.number;
```
karma-o, line 14: imagine that we have 50 more fields that describe a ball, it would be quite challenging to understand what is what if we describe it like that, so we need to add an indication of what field has what value, and make it JSON formatted to make it even prettier, so I suggest using the default format that we have in code generation feature of the IDE:
```suggestion
        return "Ball{"
                + "color=" + color
                + ", number=" + number
                + '}';
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
     public String getRandomColor() {
-        return null;
+        return String.valueOf(Colors.values()[new Random().nextInt(Colors.values().length)]);
```
karma-o, line 8: Let's split it up a bit first because it is not readable at the moment.
Then we make random a class field to avoid creating a new one each time
After that, we can add a variable for `Color.values()` call result to avoid multiple calls

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
karma-o, line 3: enum class names should be in singular form as per the style guide

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        String newBallColor = new ColorSupplier().getRandomColor();
+        int newBallNumber = new Random().nextInt(100);
```
karma-o, line 8: `Random` and `ColorSupplier` should be a class level fields, plus naming is incorrect, there variables are not color and number yet, they are a services which provide us with some values, and should be named properly

karma-o, line 8: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99
And make it a constant

karma-o (changes requested):
Check common mistakes before sending your solution to review, next time it won't be reviewed until those mistakes are fixed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private Random randomNumberService = new Random();
+    private final ColorSupplier colorSupplierService = new ColorSupplier();
```
karma-o, line 7: I'd say that now your field names are a bit too verbose, it is fine if we name it just as they are and do not add anything else, I think we can call it a good practice to name classes like this just by their name, otherwise it may cause confusion
```suggestion
    private Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random randomNumberSupplier = new Random();
```
karma-o, line 6: same here

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplierService.getRandomColor(), randomNumberService.nextInt(101));
```
karma-o, line 10: 101 is magic number, it should be constant, as I said in the previous review

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 7: 3 can also be a constant

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
...
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball.toString());
```
karma-o, line 9: can we inline 2 operations and remove that `ball` variable?
Also `toString()` is not necessary to call if you pass it as a method parameter

karma-o (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int randomNumberVerge = 101;
```
andrii-hoienko, line 6: This field is constant and should be named properly, not in camel case

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        int sizeOfEnum = Color.values().length;
+        int indexOfReturnedColor = random.nextInt(sizeOfEnum);
```
andrii-hoienko, line 10:
```suggestion
        int index = random.nextInt(sizeOfEnum);
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        int sizeOfEnum = Color.values().length;
```
andrii-hoienko, line 9:
```suggestion
        int colorsLength= Color.values().length;
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int countOfIterations = 3;
```
andrii-hoienko, line 4: Same here, regarding naming

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    Ball(Color color, int number) {
```
andrii-hoienko, line 7: add access modifier

andrii-hoienko (changes requested):
Good job, left minor comments

andrii-hoienko (approved):
Good job!

PR #1309, opened 2024-01-04
run with error

vladshender:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
okuzan, line 3: according to enum naming convention, we must use singular - Color

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery game = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(game);
```
okuzan, line 7:
```suggestion
            System.out.println(game.getRandomBall());
```
have you tested your program? :) We need to print the result of the getRandomBall method call, not just lottery, which doesn't even have toString overrided to be human readable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
okuzan, line 7: let's make Random a class-level private field, to avoid recreating it multiple times needlessly

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
```
okuzan, line 8: same for Random and ColorSupplier here - extract to fields

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        Ball bullet = new Ball();
+        bullet.setNumber(random.nextInt(100));
+        bullet.setColor(color.getRandomColor());
```
okuzan, line 11: Create a constructor in the Ball class that can accept these two arguments.
Make 101 a class constant, use CONSTANT_CASE (101 if we want to follow the requirements precisely, now the maximum possible number will be 99, look up how nextInt works - passed value is excluded)

pom.xml
```diff
@@ -9,7 +9,7 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <jdk.version>17</jdk.version>
+        <jdk.version>11</jdk.version>
```
okuzan, line 12: what's the reason for this change? If you have a problem with Java versions, don't hesitate to ask for help in our chat!

okuzan (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
+package core.basesyntax;
+
+public class Ball {
+    public static final int MAX_NUMBER_BALL = 101;
```
okuzan, line 4: ball model shouldn't be concerned with issues of Lottery implementation like this - move it to Lottery class, and there you can make it private

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        return Color.values()[index].toString();
```
okuzan, line 10: [Checklist ](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery)issue :) read about toString vs name() for enums

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery game = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(game.getRandomBall().toString());
```
okuzan, line 7:
```suggestion
            System.out.println(game.getRandomBall());
```
toString will be called under the hood

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class Lottery {
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+    private int value = Ball.MAX_NUMBER_BALL;
+
+    public int getValue() {
+        return value;
+    }
+
```
okuzan, line 13:
```suggestion
```

okuzan (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+    private static final int MAX_NUMBER_BALL = 101;
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+
+    public String getRandomBall() {
+        int num = random.nextInt(MAX_NUMBER_BALL);
+        String col = color.getRandomColor();
+        return num + " " + col;
+    }
+
```
okuzan, line 15:
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    private static final int MAX_NUMBER_BALL = 101;
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+
+    public String getRandomBall() {
+        int num = random.nextInt(MAX_NUMBER_BALL);
+        String col = color.getRandomColor();
+        return num + " " + col;
+    }
```
okuzan, line 14: here, you should create a Ball object
Ball class needs toString method overrided - which will be used to get String representation of Ball object, instead of constructing that string here
Also, be mindful of informative naming - you don't save much time not typing out a few extra letters, but that makes code harder to read for others. Remember, that we spend more times reading others code, than writing our own :)
Change to number, color, use a ball constructor

okuzan (changes requested):

vladshender:
thanks for your patience!

src/main/java/core/basesyntax/Lottery.java
okuzan: not fixed

okuzan (approved):
Well done 😎 🚀

PR #1310, opened 2024-01-05
Main method create tree random balls

OleksandrStrilets:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
+        Ball ball1 = new Lottery().getRandomBall();
+        Ball ball2 = new Lottery().getRandomBall();
+        Ball ball3 = new Lottery().getRandomBall();
+        System.out.println(ball1.toString());
+        System.out.println(ball2.toString());
+        System.out.println(ball3.toString());
```
olekskov, line 10: Please use a loop for repetitive actions
https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-for-loop-for-creating-several-objects-of-the-same-class

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: Class names should be in singular form
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier randomColor = new ColorSupplier();
+        Ball ball = new Ball(new Random().nextInt(100), randomColor.getRandomColor());
```
olekskov, line 8: 1. Please make ColorSupplier and Random class variable to avoid creation of new object each time method is called
https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level

2. Please extract magic number to class constant with proper naming
https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
olekskov, line 7: Random -> class variable
https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private Colors color;
+
+    public Ball(int number, Colors color) {
+        this.number = number;
+        this.color = color;
+    }
+
+    @Override
+    public String toString() {
+        return number + " " + color.toString();
```
olekskov, line 14: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants
```suggestion
        return number + " " + color.name();
```

olekskov (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
+        for (int i = 0; i < 3; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
Ivan95kos, line 6: create one instance of `Lottery`
```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random random = new Random();
+    private int index = random.nextInt(Color.values().length);
+
+    public Color getRandomColor() {
+        return Color.values()[index];
+
     }
```
Ivan95kos, line 12:
```suggestion
    public Color getRandomColor() {
        return Color.values()[random.nextInt(Color.values().length)];
    }
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier randomColor = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(new Random().nextInt(100), randomColor.getRandomColor());
```
Ivan95kos, line 9: `Random` must be a class field

Ivan95kos (changes requested):
left few comments

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    static final int NUMBER = 100;
+    private ColorSupplier randomColor = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball(new Random().nextInt(NUMBER), randomColor.getRandomColor());
+        return ball;
```
okuzan, line 11:
```suggestion
        return new Ball(new Random().nextInt(NUMBER), randomColor.getRandomColor());
```
In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass the expression directly without assigning it to a variable

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 9: you can remove the comments if you don't need them

okuzan (approved):
well done

PR #1314, opened 2024-01-06
The task has been solved

GosudarBBV:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color="
+                + color
```
karma-o, line 36:
```suggestion
                + "color=" + color
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+
+    private static final Color[] colors = Color.values();
+    private static final Random random = new Random();
+
+    public static Color getRandomColor() {
```
karma-o, line 10: Let's just avoid static methods and variables for now(except for constant values like numbers) and reformat our solution by OOP style, which will be covered in detail later in the course.
Same for all other classes and their methods

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+
+    private static final Color[] colors = Color.values();
```
karma-o, line 7: this array can be a local variable inside a method. 
Usually, class-level variables are some services or fields that are part of a bigger entity, like engine is a part of a car, or an email is a detail of some user, but an array like this is more of a part of a method and is not worth to be stored in a class, even if it is a constant value like in this case

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
```
karma-o, line 4: next time try to keep the original method signature, since most of the time it defines a way how a task should have been done in the first place

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
+
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(100));
```
karma-o, line 9: Do not create a new `Random` instance each time, make it a class-level variable.
Also, check the documentation of the nextInt() method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public static Ball getRandomBall() {
+
```
karma-o, line 8: redundant empty line
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(100));
+    }
+}
+
```
karma-o, line 12:
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,15 @@
...
-        return null;
+
+    private static final Color[] colors = Color.values();
+    private static final Random random = new Random();
+
+    public static Color getRandomColor() {
+
```
karma-o, line 11:
```suggestion
```

karma-o (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final Lottery lottery = new Lottery();
+        Ball[] balls = {lottery.getRandomBall(), lottery.getRandomBall(),
+                lottery.getRandomBall(), lottery.getRandomBall()};
```
kerrrusha, line 8: 1. storing balls is not neccesary, as we dont want to use them later. 
2. what if we would want to create 100500 balls? create balls in loop instead
```suggestion
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,39 @@
+package core.basesyntax;
+
+public class Ball {
+
```
kerrrusha, line 4:
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
kerrrusha: not fixed

src/main/java/core/basesyntax/Lottery.java
kerrrusha: not fixed

src/main/java/core/basesyntax/ColorSupplier.java
kerrrusha: not fixed

src/main/java/core/basesyntax/ColorSupplier.java
kerrrusha: not fixed

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,39 @@
...
+    private Color color;
+    private int number;
+
+    public Ball() {
+    }
```
kerrrusha, line 9: KISS/YAGNI principle - we dont need it
```suggestion
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
...
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
```
artemZhuravskyi, line 26:
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
artemZhuravskyi: not fixed

src/main/java/core/basesyntax/Lottery.java
artemZhuravskyi: fixed

src/main/java/core/basesyntax/Lottery.java
artemZhuravskyi: fixed

artemZhuravskyi (changes requested):
Good job. Don't forget to fix previous comments.

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final Lottery lottery = new Lottery();
+        Ball[] balls = lottery.createBalls(5);
```
kerrrusha, line 7: > To test the program, in the main() method, create three balls using the getRandomBall() method from the Lottery

src/main/java/core/basesyntax/Lottery.java
kerrrusha: no, its not
```suggestion
```

kerrrusha (approved):
this a little fixes, hope you can fix them by yourself

PR #1316, opened 2024-01-07
Creating lottery random balls

Svitlankaa:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
Ivan95kos, line 7: Random must be a class field
```suggestion
        int index = random.nextInt(Color.values().length);
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Ivan95kos, line 8: `Random `and `ColorSupplier `must be a class field

Ivan95kos (changes requested):
Good job overall!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
```
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
+    private static final int NUMBER_RANDOM = 100;
```
okuzan, line 7:
```suggestion
    private static final int NUMBER_RANDOM = 101;
```
better use 101 because the value we pass to nextInt is excluded from possible values

okuzan (approved):
Well done!

PR #1318, opened 2024-01-07
Have done new classes (Lottery.java, Color.java, Ball.java), realizat…

bilberryyy:
…ion methods like getRandomColor() and getRandomNumber() and print information in console abt lottery. TY!

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        // creating three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = Lottery.getRandomBall();
+            System.out.println(ball);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        Color[] colors = Color.values();
+        int index = new Random().nextInt(colors.length);
+        return colors[index].name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        String randomColor = ColorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(101);
+        return new Ball(randomColor, randomNumber);
+    }
+}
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        String randomColor = ColorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(101);
+        return new Ball(randomColor, randomNumber);
+    }
```
okuzan, line 10: let's make color supplier, random - private class-level fields, and 101 - private class-level constant (don't forget to use CONSTANT_CASE, name it smth like BOUND). Good insight for choosing 101 :)
We should avoid static methods, as mentioned in the [checklist](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
```
okuzan, line 6: remove static

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Color[] colors = Color.values();
+        int index = new Random().nextInt(colors.length);
```
okuzan, line 8: make Random a field and use it here. This way we make our code more efficient - no need to recreate random instance per each method call

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        // creating three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = Lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
okuzan, line 9: In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable (ball).
Create one lottery instance before the iteration(we will need it bcuz it won't be static)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        // creating three balls using class Lottery and print information about them in console
```
okuzan, line 5: you can remove the comments if you don't need them

okuzan (changes requested):
Good job on your first try!

okuzan (changes requested):
Make sure to re-request a review after you have pushed your new changes so that your homework is added to the queue on our platform, I came across your homework accidentally this time :)

bilberryyy:
Hi, i'm so sorry. May you please confirm whether the issue has been
addressed and if my revised HW is in the line?

вт, 9 янв. 2024 г. в 14:05, Oleh Kuzan ***@***.***>:

PR #1320, opened 2024-01-08
The task has done

NekitCool1:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int constant = 1000;
```
sokimaaa, line 6: Change to the meaningful constant name
```suggestion
    private static final int <CHANGE_TO_PROPER_CONSTAN_NAME>= 1000;
```

sokimaaa (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int CONSTANT = 1000;
```
sokimaaa, line 6: constants should have meaningful (aka self descriptive) name

sokimaaa (changes requested):
constants should have meaningful (aka self descriptive) name

PR #1325, opened 2024-01-16
Hw 3 solution

ESK86:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int lottery_count = 3;
+        for (int i = 0; i < lottery_count; i++) {
+            Lottery lot = new Lottery();
```
kerrrusha, line 8: create lottery outside the loop. we dont need to create new lottery each time we want to get ball

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,10 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        final int lottery_count = 3;
```
kerrrusha, line 6: make it `private static final` class-level field with correct letter case

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
kerrrusha, line 3: should be in singular not plural
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        return new Ball(random.nextInt(100));
```
kerrrusha, line 8: magic number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
```
kerrrusha, line 7: should be class-level field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
kerrrusha, line 7: random should be class-level field

kerrrusha (changes requested):

PR #1327, opened 2024-01-19
added 'Lottery' class which returns instance of 'Ball' class with …

DanilKi:
…a random state ('color' & 'number' fields)

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import core.basesyntax.model.Ball;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import core.basesyntax.model.Color;
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random randomColor = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = randomColor.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+        return color.name().toLowerCase();
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+import core.basesyntax.model.Ball;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
```
okuzan, line 12: don't create array in this class, print the result of method call directly

src/main/java/core/basesyntax/model/Ball.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax.model;
+
+public class Ball {
+    private String color;
+    private int number;
```
okuzan, line 5: let's create constructor for these fields

okuzan (changes requested):
Very good job on the first try! 🔥

src/main/java/core/basesyntax/model/Ball.java
DanilKi: added that constructor, and changed getRandomBall() method

src/main/java/core/basesyntax/Application.java
DanilKi: removed creation of array, now main() metod prints return of getRandomBall() method directly

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import core.basesyntax.model.Color;
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random randomColor = new Random();
```
kerrrusha, line 7: random number generator has no common with color
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import core.basesyntax.model.Ball;
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random randomNumber = new Random();
```
kerrrusha, line 9: it's not random number but random generator, better just name it as `random`

kerrrusha (approved):

src/main/java/core/basesyntax/Lottery.java
DanilKi: Ok

src/main/java/core/basesyntax/ColorSupplier.java
DanilKi: Ok, clear

PR #1329, opened 2024-01-21
Home work solution for creating lottery ball

k0sm0naft:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+    private static final Lottery LOTTERY = new Lottery();
```
karma-o, line 5: It may look like a constant because we initialize it only once here, but this is essentially a service that is usually initialized in the constructor and created somewhere outside of this class to not hardcode them (google `Dependency injection` and `Inversion of control` if you are curious but do not dig too deep since it is covered in details in the course)

So it can not be named as a constant, usual `lottery` would be fine, and lets not make it static, to follow-up with the OOP principles :)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    @Override
+    public String toString() {
+        String additionalZero = "";
+        if (number < 10) {
+            additionalZero += 0;
+        }
+        return "Ball: " + additionalZero + number + " " + color;
+    }
```
karma-o, line 19: All of that should not be a part of the `Ball` class logic, `toString()` should just give us a string representation of that class, without any counters, `if`s or anything basically, if you want to do this in a pretty way, just add some more logic to the main method.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int randomIndex = random.nextInt(Colors.values().length);
+        return Colors.values()[randomIndex].name();
```
karma-o, line 10: `Colors.values()` call result can be saved to a variable to avoid multiple calls

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
karma-o, line 3: enum class names should be in the singular form as per the style guide
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER_ON_BALL = 100;
```
karma-o, line 6: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99

karma-o (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,22 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Application application = new Application();
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            application.printBall(lottery.getRandomBall());
+        }
+    }
+
+    private void printBall(Ball ball) {
```
karma-o, line 14: I'd say that for the Main class we can make an exception and make it static to avoid creating an `Application` class instance

karma-o (approved):
GJ! Left a minor comment to the main class

src/main/java/core/basesyntax/Application.java
k0sm0naft: I'll keep that in mind.

PR #1331, opened 2024-01-23
task done

SukhachovS:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
okuzan, line 10: create one lottery instance
call its method in the loop

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,35 @@
+package core.basesyntax;
+
+public class Ball {
+
```
okuzan, line 4:
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static String getRandomColor() {
+        Random random = new Random();
```
okuzan, line 7: make Random a class field , we don't need to recreate it for each method call
we should avoid static methods

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        Random random = new Random();
```
okuzan, line 7: random shouldn't be a local variable, but a class field, same for color supplier
make 100 a class constant, use CONSTANT_CASE for naming, don't forget appropriate access modifier for all of these

okuzan (changes requested):
Good job! Please always make sure to review the points mentioned in the [checklist ](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md)of common mistake before submitting your solution. It's located at the end of the task description

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Ball ball;
+        for (int i = 0; i < 3; i++) {
+            ball = new Lottery().getRandomBall();
+            System.out.println(ball);
+        }
```
olekskov, line 9: 1. we should move out Lottery creation out of loop as we need only one object of it
2. variable for ball is not necessary
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < 3; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

olekskov (approved):
Good job, approved.
Please check comment below to avoid same in the future

PR #1335, opened 2024-01-24
main example

EduarfF:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier randomColor = new ColorSupplier();
```
karma-o, line 7: this is not a color yet
```suggestion
        ColorSupplier colorSupplier = new ColorSupplier();
```

karma-o (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Color[] COLORS = Color.values();
+    private static final Random RANDOM = new Random();
```
karma-o, line 7: Also about the naming of `Random` var, it may look like a constant because it is static and final here, but usually objects like that are dependencies that are initialized in the constructor or outside of that class. (which you can also do here)
On the other hand, constants are usually some primitives/wrappers/arrays or other data structures of things, etc.
So it can be named as just `random` and initialized in the constructor

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
...
+    private ColorSupplier colorSupplier;
+    private Random random;
+
+    public Lottery() {
+        this.colorSupplier = new ColorSupplier();
+        this.random = new Random();
+    }
+
+    public Ball getRandomBall() {
+        final int randomNumber = random.nextInt(101);
```
karma-o, line 15: 101 should be a constant as I mentioned in the first review

karma-o (changes requested):
GJ! Added jsut a couple of clarifications

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -3,16 +3,10 @@
...
 public class ColorSupplier {
-    private Random random;
-
-    public ColorSupplier() {
-        random = new Random();
-    }
-
     public String getRandomColor() {
         Color[] colors = Color.values();
+        Random random = new Random();
```
karma-o, line 8: Here we basically got to what we started with, this `random` variable should still be a field of `ColorSupplier` as Yevhenii said in his comment, to avoid its multiple instances creations, but he also pointed out that it should be initialized in the field right away and not in the constructor. 

I am sorry if the comments are confusing, I personally do not fully agree with the constructor detail of the review and do not see that big of a difference, but anyway, do not be afraid to tell us if you see a contradiction in the mentor comments, we try to avoid it, but sometimes they just happen. You can ask us both about it in the personal chat anytime and all of us can come up with the best possible approach if we discuss it together :)
We can still do that if you'd like, but for now, to avoid spending even more time on this, I'd rather agree with Yevhenii's comment, so let's do as he proposed

karma-o (changes requested):
Left a comment on the changes and the review itself

PR #1339, opened 2024-01-26
Solution of jv-lottery

makaroshyna:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public class Ball {
+
+    private final String color;
+    private final int number;
```
karma-o, line 6: I like this nice touch of making those fields final. Makes total sense here, GJ! :+1: 
Just be careful with that in the future, because usually, our POJO classes like this are mutable
You can read more about it [here](https://www.baeldung.com/java-immutable-object)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index].name();
```
karma-o, line 8: do not create a new random each time, make it a class field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random number = new Random();
```
karma-o, line 9: naming is important, this is not a number, but a class that can provide us with all sorts of random values. Just `random` would be fine

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
```
karma-o, line 6: check the documentation of the `nextInt()` method, bound parameter there is exclusive, which means that right now the maximum possible number that we can get is 99

karma-o (changes requested):

makaroshyna:
made all the changes you suggested 👌🏼

PR #1340, opened 2024-01-26
Hw jv lottery

mykolazx3:


karma-o (changes requested):
Please check common mistakes in the task description before sending your solution to review, next time it won't be reviewed until the mistakes described there are fixed

mykolazx3:
fixed

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
+
```
karma-o, line 11: redundant empty lines: 7 and 11

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
+    public static final int AMOUNT_OF_COLORS = Color.values().length;
```
karma-o, line 7: I do not see a reason in having it as a constant, constants are usually some custom things and values that are not clear why they are where they are if they would not be named. Let's take a closer look at `100` that you had, it is not very readable if it's just plain `100` in the code, but if you name it as a `max number` or something like that - it becomes obvious.

so let's just put it back as a method variable

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 8: 3 can be a constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static final int AMOUNT_OF_BALL = 99;
```
karma-o, line 6: Few things here:
* naming - amount of ball can be translated more like "кількість м'ячів" but what it should mean is "максимальне число на м'ячі"
* I left a comment on the exclusivity of that number, so lets think what does exclusive number means in the context of the range of numbers, when for example I say we have a range from inclusive 2 up to 5 exclusive, let's list numbers in that range in the comment here, and also fix a problem that we have in the code. You can also check out the Javadoc of the `nextint()` method 😉

karma-o (changes requested):

karma-o:
also a very important thing, I can see that you are trying to fix comments directly on github, It can be useful sometimes when all you need to do is like 1-2 lines of code, but in general it is a bad habbit for a number of reasons: 
* commit history becomes a mess 
* you can't see any compilation errors 
* you have to pull the changes that you have made here from git to the IDE 
* tests can fail often because you do not run them locally before push = even bigger mess in commits from small changes
So let's avoid that in the future

src/main/java/core/basesyntax/Lottery.java
mykolazx3: should I provide here amount from specific range from x to y? (instead from 0 to 98)

PR #1341, opened 2024-01-27
Lottery homework

kasiashka:
created classes, implemented method getRandomColor(), realized getRandomBall() method

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+        System.out.println(Arrays.toString(balls));
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+        return color.name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Random random = new Random();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
+        return ball;
+    }
+}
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
sokimaaa, line 7: random should be instantiated once and reused multiple times 

consider making it class level field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Random random = new Random();
```
sokimaaa, line 9: random should be instantiated once and reused multiple times

consider making it class level field
```suggestion
        Random random = new Random();

```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Random random = new Random();
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
sokimaaa, line 12: all magic numbers should be constant with self descriptive name
```suggestion
        ball.setNumber(random.nextInt(CONSTANT));

```

sokimaaa (changes requested):

kasiashka (commented):
rework mistakes

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private final int maxBallNumber = 100;
```
okuzan, line 8: constants are static
use constant case

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
```
okuzan, line 9: rewrite main to not use arrays , we can print the result of method call directly

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
```
okuzan, line 11: call print here

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+    private final int maxBallNumber = 100;
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(maxBallNumber));
+        return ball;
```
okuzan, line 14: why not create a constructor for these fields?

okuzan (changes requested):

Rommelua (approved):

PR #1342, opened 2024-01-27
First Solution, added 3 new classes and modified 2

kacper-kukula:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum name should be in singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), random.nextInt(101));
```
okuzan, line 10: let's extract 101 to a class-level constant

okuzan (changes requested):
Very good job on your first try!

Rommelua (approved):

PR #1349, opened 2024-02-03
After review solution

esidarapp:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
```
```suggestion
    private final Random random = new Random();
```
do the same with color supplier

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+public class Lottery {
+    private static final Random random = new Random();
+    private static final int BOUND = 100;
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        String color = colorSupplier.getRandomColor();
+        int num = random.nextInt(BOUND);
+        Ball ball = new Ball(color, num);
+        return ball;
```
okuzan, line 14: in such cases, don't store to a local variable, use expression right away
```suggestion
        return new Ball(color, num);
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
```suggestion
            System.out.println(lottery.getRandomBall());
        }
```
same here

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Lottery lottery = new Lottery();
```
okuzan, line 6: move lottery creation out of the loop, why create 3 lottery instances, when you only need one?

okuzan (changes requested):

PR #1351, opened 2024-02-03
All classes were added, all methods were implemented by the task. Bui…

slizko1:
All classes were added, all methods were implemented by the task. Build successful!

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,11 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        Lotery lotery = new Lotery();
+        for (int i = 0; i < 3; i++) {
+            balls[i] = lotery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].name();
     }
 }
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[3];
+        Lotery lotery = new Lotery();
+        for (int i = 0; i < 3; i++) {
+            balls[i] = lotery.getRandomBall();
+            System.out.println(balls[i].toString());
+        }
```
karma-o, line 11: 3 is a magic number, make it a constant
Also, I think we can remove the array and just generate a ball and print it right away, we can even remove the `toString()` call, since it will be called automatically if we just pass that ball to sout like that:
```java
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
karma-o, line 3: enum class names should be singular form as per [style guide](https://google.github.io/styleguide/javaguide.html#s4.8.1-enum-classes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
karma-o, line 7: can we reuse a single instance of Random instead of creating a new one each time when `getRandomColor()` is called?
check common mistakes for answer

src/main/java/core/basesyntax/Lotery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lotery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Ball ball = new Ball();
+        ball.setNumber(new Random().nextInt(100));
```
karma-o, line 9: same here for `ColorSupplier` and `Random`
Also 100 should be constant

karma-o (changes requested):
Please check common mistakes in the task description before sending your solutions to review, next time it won't be reviewed until the mistakes described there are fixed

PR #1353, opened 2024-02-04
Initial commit

fedaktaras:
It seems to be okay, but there is a chance to get the same color and number for the ball two or even three times in a row.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,9 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println("First " + lottery.getRandomBall());
+        System.out.println("Second " + lottery.getRandomBall());
+        System.out.println("Third " + lottery.getRandomBall());
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,10 +1,11 @@
 package core.basesyntax;
 
 public class Application {
+    private static final Lottery lottery = new Lottery();
+    private static final int numberOfRandomBalls = 3;
+
     public static void main(String[] args) {
-        Lottery lottery = new Lottery();
-        final int numberOfRandomBalls = 3;
-        for (int i = 1; i <= numberOfRandomBalls; ++i) {
+        for (int i = 1; i <= numberOfRandomBalls; i++) {
             System.out.println(i + " " + lottery.getRandomBall());
         }
     }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,11 +1,12 @@
 package core.basesyntax;
 
 public class Application {
-    private static final Lottery lottery = new Lottery();
-    private static final int numberOfRandomBalls = 3;
+
+    private static final int NUMBER_OF_RANDOM_BALLS = 3;
 
     public static void main(String[] args) {
-        for (int i = 1; i <= numberOfRandomBalls; i++) {
+        Lottery lottery = new Lottery();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,9 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println("First " + lottery.getRandomBall());
+        System.out.println("Second " + lottery.getRandomBall());
+        System.out.println("Third " + lottery.getRandomBall());
```
karma-o, line 9: let's generate and sout balls in a loop

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
karma-o, line 7: make random a class-level var

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Color someColor = Color.values()[random.nextInt(Color.values().length)];
+        return someColor.toString();
```
karma-o, line 9: use enum's `name()` instead of `toString()` try to think why it would be better 😉

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
```
karma-o, line 9: let's also avoid the creation of new `ColorSupplier` and `Random` each time this method is called

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        Random random = new Random();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
karma-o, line 11: magic number should be a constant
Also a bound parameter of `nextInt()` is exclusive, so we should account for that, check Javadoc of the `nextInt()` for details

karma-o (changes requested):
Please check common mistakes in the task description before submitting your solution for review, next time it won't be checked until they are fixed

src/main/java/core/basesyntax/ColorSupplier.java
fedaktaras: toString() can be overridden name() is final and it is ensuring that it will always return the exact name of the enum constant as declared in the enum type. Better to read code. The explanations I found.

yevheniisynytsia (changes requested):
Well done! Left some comments to work on

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
 public class ColorSupplier {
+    private final Random random;
+
+    public ColorSupplier() {
+        random = new Random();
+    }
+
     public String getRandomColor() {
-        return null;
+        Color someColor = Color.values()[random.nextInt(Color.values().length)];
```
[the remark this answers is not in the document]
fedaktaras: Agree. It is much better.

andrii-hoienko (changes requested):
Nice, but some improvements required

andrii-hoienko (approved):
Gj

PR #1355, opened 2024-02-05
Resolving jv-lottery task

DmitryLazyuka:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(), new Random().nextInt(101));
```
olekskov, line 9: Please extract random to class variable same as done for ColorSupplier in order not to create new object each time method is called

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
olekskov, line 7: same here

src/main/java/core/basesyntax/Lottery.java
olekskov, line 9: Please extract magic number to constant with proper naming

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
+        return Color.values()[index].toString();
```
olekskov, line 8: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants
```suggestion
        return Color.values()[index].name();
```

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random RANDOM_NUMBER = new Random();
```
olekskov, line 6: This is not a number actually. It's special service that provides random numbers, naming should represent it
```suggestion
    private final Random randomProvider = new Random();
```
or
```suggestion
    private final Random random = new Random();
```
or
```suggestion
    private final Random randomSupplier = new Random();
```
or anything like that

olekskov (approved):
Good job, approved.
Please pay attention to naming for future

PR #1358, opened 2024-02-07
added new classes and methods

AnyaRatusnyak:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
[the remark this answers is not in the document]
AnyaRatusnyak: Thank you) Yes , I can create a single instance of Random for the entire class .it reduces the overhead of object creation and initialization. Reusing the same Random instance ensures consistent behavior across multiple method invocations. The Random class is not thread-safe, meaning it may produce unpredictable behavior if accessed concurrently by multiple threads.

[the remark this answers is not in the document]
karma-o: I guess thread safety would be a problem for another day then :)
But we talk about it seriously, usually applications are either fully multhithreading oriented, safe and designed with a thought about it from the beginning of the development, or just a simpler apps that can partially use some parallelism but that do not support that fully and should be refactored if we want to accommodate for that.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -3,10 +3,12 @@
...
 
 public class Lottery {
+    private static final Random random = new Random();
+    private ColorSupplier supplier = new ColorSupplier();
+
     public Ball getRandomBall() {
-        ColorSupplier supplier = new ColorSupplier();
         String color = supplier.getRandomColor();
-        int number = new Random().nextInt(100);
+        int number = random.nextInt(100);
```
karma-o, line 11: magic number should be a constant
Also, a bound parameter of `nextInt()` is exclusive, so we should account for that, check Javadoc of the nextInt() for details

src/main/java/core/basesyntax/Application.java
```diff
@@ -4,8 +4,8 @@ public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
         Lottery lottery = new Lottery();
-        System.out.println(lottery.getRandomBall().toString());
-        System.out.println(lottery.getRandomBall().toString());
-        System.out.println(lottery.getRandomBall().toString());
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 7: 3 can also be a constant here

src/main/java/core/basesyntax/Application.java
```diff
@@ -4,8 +4,8 @@ public class Application {
...
+            System.out.println(lottery.getRandomBall().toString());
```
karma-o, line 8: `toString()` call is not necessary, because sout recieves a param of type string and will make that call automatically

karma-o (changes requested):

andrii-hoienko (approved):
Good job!

PR #1361, opened 2024-02-08
Create a simple lottery that will return balls with a random color an…

VArabchuk:
…d random number

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,8 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
+        new Lottery().getRandomBall();
+
         // create three balls using class Lottery and print information about them in console
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,18 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+        this.color = new ColorSupplier().getRandomColor();
+        this.number = new Random().nextInt(100);
+    }
+
+    @Override
+    public String toString() {
+        return "Ball number " + number + " is " + color;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors color = Colors.values()[new Random().nextInt(Colors.values().length)];
+        return color.toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
+    BLUE,
+    RED,
+    GREEN,
+    ORANGE
+
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,9 @@
+package core.basesyntax;
+
+public class Lottery {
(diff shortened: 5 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
+        new Lottery().getRandomBall();
```
okuzan, line 5: call this method in a loop, but using only one lottery instance

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
+    BLUE,
+    RED,
+    GREEN,
+    ORANGE
+
```
okuzan, line 8:
```suggestion
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum should always be in singular

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Random;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+        this.color = new ColorSupplier().getRandomColor();
+        this.number = new Random().nextInt(100);
+    }
```
okuzan, line 12: separation of concerns design principle - ball model itself shouldn't be bothered with logic of generating random properties, it should be independent. This logic should be in lottery class instead. Ball should allow its users to create a ball given some values. Why limit yourself to the only possible usage scenario? Create a constructor which takes two arguments

read where `this` is needed - https://stackoverflow.com/a/2411283

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors color = Colors.values()[new Random().nextInt(Colors.values().length)];
+        return color.toString();
```
okuzan, line 8: image
https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery
Make random a class-level field instead, to avoid needless recreation per each method execution

okuzan (changes requested):
Please always make sure to read and fix the common mistakes checklist before requesting a PR. Good job so far!

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Color color = Color.values()[new Random().nextInt(Color.values().length)];
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Rommelua, line 7: image

image

Rommelua (changes requested):

Rommelua (changes requested):
see comments

src/main/java/core/basesyntax/Lottery.java
VArabchuk: The provided link is not working, please provide a working one.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        int counter = 0;
+        while (counter < 3) {
+            System.out.println(lottery.getRandomBall());
+            counter++;
+        }
```
okuzan, line 10: for loop is more suitable for such cases

okuzan (approved):
Good job

PR #1366, opened 2024-02-10
Implemented the task solution

ghost:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Random random = new Random();
+    private final Color[] colors = Color.values();
+    private final int maxIndex = colors.length;
```
karma-o, line 8: I think these vars are too method-specific to be class variables, so let's move them

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
karma-o, line 6: 3 can be a constant here

karma-o (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
VlasenkoYevhenii: I would argue about how specific these variables are because there would be excessive  method/field calls every time 
 the getRandomColor() method is being called.

yevheniisynytsia (approved):
Left my thoughts about that variable. Everything else looks good! Well done!

PR #1370, opened 2024-02-13
Created simple lottery!

Recdt:


karma-o (changes requested):
check and fix common mistakes in the task description before submitting your solution

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int colorNumber = new Random().nextInt(Color.values().length);
+        return Color.values()[colorNumber].toString();
```
karma-o, line 8: another point from the checklist
```suggestion
        return Color.values()[colorNumber].name();
```

karma-o (changes requested):
almost nothing from common mistakes was fixed, you should be more attentive with that

PR #1371, opened 2024-02-14
Lottery random practice

OleksiiCherniienko:


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,22 @@
...
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                +
+                "color='" + color + '\''
+                +
+                ", number=" + number
+                +
+                '}';
```
Vasul-Pankiv, line 20: bad code style, do not leave single + in row

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+        Colors randomColor = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(100);
```
Vasul-Pankiv, line 11: make magic number a constant

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: enum names should be in singular form
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+class Ball {
+    private final Colors color;
```
olekskov, line 4: Because in any place of usage it's confusing when single value of enum has type name is in multiple form
```suggestion
    private final Color color;
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
 
-public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+import java.util.Random;
+
+class ColorSupplier {
+
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
olekskov, line 8: Random should be class variable, same as in Lottery class. as we don't need to create new object each time method is called

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int number;
+
+    public Ball(Colors color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{" + "color='" + color + '\'' + ", number=" + number + '}';
```
olekskov, line 14:
```suggestion
        return "Ball{color='" + color.name() + "', number=" + number + '}';
```
or
```suggestion
        return String.format("Ball{color='%s', number='%s'", color.name(), number}';
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
```
olekskov, line 5: no need of empty line in the beginning of a method
```suggestion
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
-public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+import java.util.Random;
+
+class ColorSupplier {
+
```
olekskov, line 6: same for beginning of class
```suggestion
```

olekskov (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println("Ball " + (i + 1) + ": " + ball);
```
okuzan, line 9: In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
-public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+import java.util.Random;
+
+class ColorSupplier {
+    private static final Random random = new Random();
```
okuzan, line 6: should not be static

okuzan (changes requested):

olekskov (approved):

PR #1379, opened 2024-02-15
THE BALL

KLICKICE:


okuzan (changes requested):
there's not actual code
Don't create a PR if build is not passing locally (build include tests by default). Run `mvn clean package` command. If you need assistance, have any questions or problems, don't hesitate to ask in the chat

PR #1381, opened 2024-02-15
added Lottery class

AnnaAAA3432:


MentorMA (changes requested):

AnnaAAA3432:
image
как это исправить?

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        Random random = new Random();
```
kerrrusha, line 8: random should be created once
https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level

kerrrusha (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 public class Application {
+    private static final int BALLS_COUNT = 10;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < BALLS_COUNT; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
+            System.out.println();
```
```suggestion
            System.out.println(lottery.getRandomBall());
```
In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
https://mate-academy.github.io/style-guides/java/java.html#s4.8.1-enum-classes

okuzan (changes requested):
Good job!

kerrrusha:
> image как это исправить?

we dont communicate here, only at the Mate platform

kerrrusha (approved):

PR #1383, opened 2024-02-19
created a lottery that will return balls with a random color and rand…

Alona-Martynova:
…om number (classes: Ball,Lottery, several methods) and tested the program in the main() method.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball [] setOfBalls = new Ball [3];
+        for (int i = 0; i < setOfBalls.length; i++) {
+            setOfBalls[i] = new Lottery().getRandomBall();
+        }
+        System.out.println(Arrays.toString(setOfBalls));
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,28 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball() {
+    }
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public enum ColorList {
+        RED,
+        ORANGE,
+        YELLOW,
+        GREEN,
+        BLUE,
+        VIOLET
+    }
+
+    @Override
+    public String toString() {
+        return "The random ball is " + color + " at number " + number;
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
-public class ColorSupplier {
+import java.util.Random;
+
+public class ColorSupplier extends Ball {
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(ColorList.values().length);
+        ColorList randomColor = ColorList.values()[index];
+
+        return randomColor.name();
     }
 }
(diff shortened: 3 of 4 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public Ball() {
+    }
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public enum ColorList {
```
olekskov, line 15: enum name should represent what single element is
```suggestion
    public enum Color {

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
 
-public class ColorSupplier {
+import java.util.Random;
+
+public class ColorSupplier extends Ball {
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(ColorList.values().length);
+        ColorList randomColor = ColorList.values()[index];
```
olekskov, line 9: Otherwise in place of usage it looks strange when you have single random color but type is whole ColorList
```suggestion
        Color randomColor = Color.values()[index];
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public Ball getRandomBall() {
+        int randomInt = new Random().nextInt(100);
+        return new Ball(new ColorSupplier().getRandomColor(),randomInt);
+    }
```
olekskov, line 9: Please extract Random and ColorSupplier to class variables.
With current implementation we will create new objects for each method call but this is not required. Single object may provide us lots of numbers or colors

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
-public class ColorSupplier {
+import java.util.Random;
+
+public class ColorSupplier extends Ball {
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(ColorList.values().length);
```
olekskov, line 8: same for Random, please put to class variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
-public class ColorSupplier {
+import java.util.Random;
+
+public class ColorSupplier extends Ball {
+
```
olekskov, line 6: unnecessary empty line after class decalaration

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public Ball getRandomBall() {
+        int randomInt = new Random().nextInt(100);
```
olekskov, line 7: Please move magic number to class constant with proper naming

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    private final int maxValue = 100;
```
```suggestion
    private static final int MAX_VALUE = 100;
```
it should be constant - static + constant case is lacking

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Arrays;
+
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball [] setOfBalls = new Ball [3];
```
```suggestion
        Ball[] setOfBalls = new Ball[3];
```
autoformat this file

try to rewrite this class without using the array
create only one lottery instance (outside of this loop), right now you create three of them
print result of the calling Lottery's method directly inside of a loop

okuzan (changes requested):
Good job so far!

Vasul-Pankiv (approved):

PR #1384, opened 2024-02-20
Homework jv-lottery

dmarmul:
added new Classes Ball, Lottery and enum Class Colours.
Implement methods:
getRandomColour() that return a random color, based on the enum values
getRandomBall() that create a ball with a random color and a random number

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,12 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        Ball[] balls = new Ball[3];
+        for (Ball ball: balls){
+            ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        Random random = new Random();
+        return colors[random.nextInt(colors.length)].toString();
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
+    White,
+    Black,
+    Red,
+    Yellow,
+    Orange,
+    Green,
+    Blue,
+    Purple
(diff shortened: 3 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        Ball[] balls = new Ball[3];
+        for (Ball ball: balls) {
+            ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
okuzan, line 11: there should be a space before and after semicolon : , use idea's formatter - Ctrl Alt L
try to rewrite this solution without using array

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        Random random = new Random();
```
okuzan, line 8: make random a private final class-level variable instead of a local one

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
+    White,
+    Black,
+    Red,
```
okuzan, line 6: https://google.github.io/styleguide/javaguide.html#s4.8.1-enum-classes
enum have different code style

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+public class Lottery {
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        Random random = new Random();
+        ColorSupplier colorSupplier = new ColorSupplier();
+        colorSupplier.getRandomColor();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
+        return ball;
+    }
```
okuzan, line 14: create a constructor for these fields (color, number) if you need them
random, supplier - same as in color supplier class, should be class fields
100 - class constant, remember, we use CONSTANT_CASE for constants

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        return colors[random.nextInt(colors.length)].toString();
```
okuzan, line 9: is toString an optimal choice here? Read common mistakes for more

okuzan (changes requested):
Always make sure to read the [common mistakes](https://mate-academy.github.io/jv-program-common-mistakes/java-core/oop/lottery) and fix them before creating a pr/ re-requesting mentor's review

PR #1385, opened 2024-02-20
Homework jv-lottery

dmarmul:
added new Classes Ball, Lottery and enum Class Colours.
Implement methods:
getRandomColour() that return a random color, based on the enum values
getRandomBall() that create a ball with a random color and a random number


dmarmul:
added new Classes Ball, Lottery and enum Class Colours.
Implement methods:
getRandomColour() that return a random color, based on the enum values
getRandomBall() that create a ball with a random color and a random number

PR #1386, opened 2024-02-21
Task complete

apgroupua:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private Random random = new Random();
+
     public String getRandomColor() {
         return null;
```
Vasul-Pankiv, line 9: Are u sure u need this?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 100;
+    private Random random = new Random();
```
Vasul-Pankiv, line 7: If u use it only one time, it not necessary to create local variable

Vasul-Pankiv (changes requested):

kozhukhovsky (changes requested):
Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the `mvn clean package` command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework

apgroupua:
My code does not compile, although it is correct. 
The compiler does not understand Color.values() from package core.basesyntax.
What should I do in this case?
Error:
Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile

okuzan (approved):
the root cause is usually at the very end
it's unreachable statement, because you haven't removed return null; we already retunred from the method
image

okuzan (changes requested):
approve misclicked

apgroupua:
Fixed it. The code compiles and runs. Thank you ...!

okuzan (approved):
Well done! 🚀

PR #1389, opened 2024-02-25
 create a simple lottery

idlt044:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
...
+        Ball[] setOfBalls = new Ball[3];
+
+        for (int i = 0; i < 3; i++) {
+            setOfBalls[i] = lottery.getRandomBall();
+            System.out.println(setOfBalls[i]);
+        }
     }
 }
+
+
```
okuzan, line 16: each file should end with one empty line, here you have three

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] setOfBalls = new Ball[3];
+
+        for (int i = 0; i < 3; i++) {
+            setOfBalls[i] = lottery.getRandomBall();
+            System.out.println(setOfBalls[i]);
+        }
```
okuzan, line 12: rewrite this solution without using an array, print the result of method call directly

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,14 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
```
okuzan, line 5: you can remove comments if you don't need them

okuzan (changes requested):
Very good job! Keep it up 💪

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        Ball[] balls = new Ball[3];
+        for (Ball ball: balls) {
+            ball = lottery.getRandomBall();
+            System.out.println(ball);
+        }
```
```suggestion
        for (int i = 0; i < 3; i++) {
            System.out.println(lottery.getRandomBall());
        }
```
Good job, but we don't really need array here, we only need to print a ball three times
format your files before committing - Ctrl Alt L - shortcut to trigger IDEA's autoformatter, there should be space before and after : (semicolon)

okuzan (changes requested):

idlt044:
Sorry my changes don`t commit and i send u wrong changes.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball);
```
```suggestion
            System.out.println(lottery.getRandomBall());
```

okuzan: In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable

okuzan (approved):
GJ 👍

PR #1398, opened 2024-03-01
Creating one enum and write code for 4 class. Program lottery complete

Wersebu:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+public class Lottery {
+    private ColorSupplier colorSupplier;
+
+    public Lottery(ColorSupplier colorSupplier) {
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Ball getRandomBall() {
+        String randomColor = colorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(100) + 1;
```
olekskov, line 14: 1. Random should be class variable, same as ColorSupplier. We don't need to create new instance for each method call
2. Magic number should be constant with proper naming

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public enum Colors {
```
olekskov, line 3: Class/Enum name should be in singular form
```suggestion
public enum Color {

```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
olekskov, line 7: Random -> class variable

olekskov (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,17 @@
...
 public class ColorSupplier {
+    private final Random random;
+
+    public ColorSupplier() {
+        this.random = new Random();
+    }
+
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
nowickimj, line 13: You've initialized the `random` field, yet you're creating a new object here.

nowickimj (changes requested):
One thing missed.

nowickimj (approved):

PR #1401, opened 2024-03-02
Make up three random balls with different enum colors in main method 

odeineka:
Create a "Ball" class with the private fields "color" and "number" and override toString method ; Create Enum with different colors; Implement "GetRandomColor" method in a "ColorSupplier" class; Create a "Lottery" class with a "getRandomBall method

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,19 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        ColorSupplier firstSupplier = new ColorSupplier();
+        ColorSupplier secondSupplier = new ColorSupplier();
+        ColorSupplier thirdSupplier = new ColorSupplier();
+        Lottery firstBall = new Lottery();
+        Lottery secondBall = new Lottery();
+        Lottery thirdBall = new Lottery();
+        System.out.println("The first number is... " + firstSupplier.getRandomColor() + " "
+                + firstBall.getRandomBall() + "!");
+        System.out.println("The second number is... " + secondSupplier.getRandomColor() + " "
+                + secondBall.getRandomBall() + "!");
+        System.out.println("The third number is... " + thirdSupplier.getRandomColor() + " "
+                + thirdBall.getRandomBall() + "!");
+
     }
 }
+
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,11 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    @Override
+    public String toString() {
+        return super.toString();
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index];
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,22 @@
+package core.basesyntax;
+
+public enum Colors {
+    violet,
+    purple,
+    magenta,
+    indigo,
+    navyBlue,
+    blue,
+    aqua,
+    green,
+    forest,
+    yellow,
+    beige,
+    orange,
+    red,
+    maroon,
+    pink,
+    brown,
+    grey,
+    black
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public int getRandomBall() {
+        Random randomBall = new Random();
+        int value = randomBall.nextInt(101);
+        return value;
+    }
+
+}


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,19 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        ColorSupplier firstSupplier = new ColorSupplier();
+        ColorSupplier secondSupplier = new ColorSupplier();
+        ColorSupplier thirdSupplier = new ColorSupplier();
```
olekskov, line 8: We can call getRandomColor() method several times from single instance of service. No need in new service to call method
```suggestion
        ColorSupplier colorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,19 @@
...
+        Lottery firstBall = new Lottery();
+        Lottery secondBall = new Lottery();
+        Lottery thirdBall = new Lottery();
```
olekskov, line 11: 1. Same as above
2. Lottery is not a ball, it's service providing balls so name ball for variable is not correct
```suggestion
        Lottery lottery = new Lottery();
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,19 @@
...
+        System.out.println("The first number is... " + firstSupplier.getRandomColor() + " "
+                + firstBall.getRandomBall() + "!");
+        System.out.println("The second number is... " + secondSupplier.getRandomColor() + " "
+                + secondBall.getRandomBall() + "!");
+        System.out.println("The third number is... " + thirdSupplier.getRandomColor() + " "
+                + thirdBall.getRandomBall() + "!");
```
olekskov, line 17: 1. Please use loop for repetative actions
2. ColorSupplier should be used in Lottery to create ball with initialized color, we don't need to use both services separately

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    @Override
+    public String toString() {
+        return super.toString();
```
olekskov, line 9: Please create your own implementation of toString() method that will print proper information instead

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,22 @@
+package core.basesyntax;
+
+public enum Colors {
+    violet,
+    purple,
```
olekskov, line 5: 1. class/enum names should be in singluar form
2. enum members are constant so should have proper NAMING_CASE
```suggestion
public enum Color {
    VIOLET,
    PURPLE,
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Colors getRandomColor() {
+        int index = new Random().nextInt(Colors.values().length);
```
olekskov, line 7: Random should be class variable as we do not need to create new instance of Random each time method is called

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+import java.util.Random;
+
+public class Lottery {
+    public int getRandomBall() {
+        Random randomBall = new Random();
+        int value = randomBall.nextInt(101);
+        return value;
+    }
+
+}
```
olekskov, line 12: 1. Lottery should provide random ball, not an int
2. Magic number should be constant with proper naming
3. This service should use both Random and ColorSupplier services to make work done
```suggestion
    public Ball getRandomBall() {
        return new Ball(random.nextInt(NUMBER_LIMIT), colorSupplier.getRandomColor().name());
    }
}
```

olekskov (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
...
         // create three balls using class Lottery and print information about them in console
+        ColorSupplier colorSupplierSupplier = new ColorSupplier();
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println("Ball" + (i + 1) + ":" + ball);
+        }
     }
 }
+
```
okuzan, line 14: only the last line of each file should be blank, you have two  here

src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        ColorSupplier colorSupplierSupplier = new ColorSupplier();
```
okuzan, line 6: remove comment and unused supplier

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index];
+
```
okuzan, line 9:
```suggestion
```
be careful with extra empty lines, review your code before committing

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 7: extract Random to a private class-level field (now it's a local var)

okuzan (changes requested):

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,24 @@
...
+        return color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    @Override
+    public String toString() {
+        return (getColor() + " " + getNumber());
```
okuzan, line 22:
```suggestion
        return getColor() + " " + getNumber();
```

src/main/java/core/basesyntax/ColorSupplier.java
okuzan: not fixed, still a local variable recreated per each method call

okuzan (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Random random = new Random();
+
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
```
okuzan, line 9:
```suggestion
    private final Random random = new Random();

    public Color getRandomColor() {
        int index = random.nextInt(Color.values().length);
```

Vasul-Pankiv (approved):

PR #1411, opened 2024-03-11
created and implemented logic for a simple lottery:

chertiav:
- create an enum Color with different colors;
- created a Ball class with private fields color and number, constructor and override the toString() method
- implemented method getRandomColor in class ColorSupplier
- created a Lottery class with the getRandomBall method
- for to test the program, in the main() method, create three balls.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final int COLOR_COUNT = Color.values().length;
+    private Random random = new Random();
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final int COLOR_COUNT = Color.values().length;
+    private Random random = new Random();
+
+    public Color getRandomColor() {
+        return Color.values()[random.nextInt(COLOR_COUNT)];
```
MentorMA, line 10: Return color in String type

MentorMA (changes requested):

Rommelua (approved):

PR #1412, opened 2024-03-11
this task is done

Tima0509:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,12 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
olekskov, line 12: Please use loop for repetitive actions

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Ball {
+    private String colour;
+    private int number;
+
+    public Ball(Color randomColor, int randomNumber) {
+    }
```
olekskov, line 8: Why do we have a constructor with unused arguments?
Please assign incoming values to object

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Color {
+    Black,
+    Blue,
+    Green,
+    Red,
+    Yellow,
```
olekskov, line 8: enum values are constants so should be spelled IN_PROPER_CASE

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
```
olekskov, line 6: According to checklist:
Don't use static methods in your solution
```suggestion
    public Color getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Random random = new Random();
```
olekskov, line 7: We do not need to create new Random object each time method is called.
Please move it to class variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        Color randomColor = ColorSupplier.getRandomColor();
+        Random random = new Random();
```
olekskov, line 8: Same here for both Random and ColorSupplier

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        int randomNumber = random.nextInt(100) + 1; // Random number between 1 and 100
```
olekskov, line 9: All magic numbers in your code should be constants.

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+public class Lottery {
+    private static ColorSupplier colorSupplier;
+
+    public Lottery(ColorSupplier colorSupplier) {
+        this.colorSupplier = colorSupplier;
+    }
+
+    public static Ball getRandomBall() {
+        Color randomColor = colorSupplier.getRandomColor();
+        Random random = new Random();
```
olekskov, line 14: Please also move Random to class variable, same as done for ColorSupplier

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+        int randomNumber = random.nextInt(100) + 1;
```
olekskov, line 15: 100 should be constant with proper naming

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static ColorSupplier colorSupplier;
+
+    public Lottery(ColorSupplier colorSupplier) {
+        this.colorSupplier = colorSupplier;
+    }
+
+    public static Ball getRandomBall() {
```
olekskov, line 12: Please do not use static methods
```suggestion
    public Ball getRandomBall() {
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball = Lottery.getRandomBall();
+            System.out.println(ball);
```
olekskov, line 7: There is no need to introduce variable for single usage when it's clear what method provides us
```suggestion
            System.out.println(lottery.getRandomBall());
```

olekskov (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier colorSupplier;
+    private Random random;
+    private final int maxRandom = 100;
```
Vasul-Pankiv, line 8: It should be constant field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int maxRandom = 100;
```
kozhukhovsky, line 6:
```suggestion
    private static final int MAX_NUMBER = 100;
```

kozhukhovsky (approved):

PR #1422, opened 2024-03-17
Created a Ball class with fields: color,number. Override toString met…

mhvaleria:
…hod.Created a Lottery which will return the ball with a random color and number.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,11 +2,9 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-
         for (int i = 0; i < 3; i++) {
-             Lottery lottery = new Lottery();
-
-             System.out.println(lottery.getRandomBall());
+            Lottery lottery = new Lottery();
+            System.out.println(lottery.getRandomBall());
         }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,7 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
+
         // create three balls using class Lottery and print information about them in console
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,7 +2,15 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
+        final int numberOfBalls = 3;
+        Ball[] balls = new Ball[numberOfBalls];
 
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < numberOfBalls; i++) {
+            balls[i] = Lottery.getRandomBall();
+        }
+
+        for (int i = 0; i < numberOfBalls; i++) {
+            System.out.println("Ball " + (i + 1) + ":" + balls[i]);
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,15 +2,11 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        final int numberOfBalls = 3;
-        Ball[] balls = new Ball[numberOfBalls];
 
-        for (int i = 0; i < numberOfBalls; i++) {
-            balls[i] = Lottery.getRandomBall();
-        }
+        for (int i = 0; i < 3; i++) {
+             Lottery lottery = new Lottery();
 
-        for (int i = 0; i < numberOfBalls; i++) {
-            System.out.println("Ball " + (i + 1) + ":" + balls[i]);
+             System.out.println(lottery.getRandomBall());
         }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,7 @@
 
 public class Application {
     public static void main(String[] args) {
+
```
okuzan, line 5: Here, you should create balls inside a for loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,26 @@
...
+    public static void main(String[] args) {
+
+        Ball ball1 = Lottery.getRandomBall();
+        Ball ball2 = Lottery.getRandomBall();
+        Ball ball3 = Lottery.getRandomBall();
+
+        System.out.println("Ball 1:" + ball1);
+        System.out.println("Ball 2:" + ball2);
+        System.out.println("Ball 3:" + ball3);
+    }
```
okuzan, line 25: should be in Application.java
naming with numbers is a sign something is wrong - here, you should do this in a loop

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public static Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(101);
```
okuzan, line 11: 101 - should be a class-level constant, random - a class-level field

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
okuzan, line 7: same for this random

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+public enum Color {
+    RED("red"),
+    GREEN("green"),
+    BLUE("blue"),
+    YELLOW("yellow"),
+    BLACK("black"),
+    WHITE("white");
```
okuzan, line 9: not a mistake but I don't see much point in having the same colorname as enum value but in lowercase, we can call .lowercase() on enum.name()

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX = 101;
```
Rommelua, line 6:
```suggestion
    private static final int MAX_BALL_NUMBER = 101;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public static Ball getRandomBall() {
```
Rommelua, line 9: image

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,14 @@
...
+    RED,
+    GREEN,
+    BLUE,
+    YELLOW,
+    BLACK,
+    WHITE;
+
+    public String getColorName() {
+        return name().toLowerCase();
+    }
```
Rommelua, line 13:
```suggestion
```

Rommelua (changes requested):

src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,16 @@
...
+
+public enum Color {
+    RED,
+    GREEN,
+    BLUE,
+    YELLOW,
+    BLACK,
+    WHITE;
+
+    public String
```
kozhukhovsky, line 11: what with formatting?

src/main/java/core/basesyntax/Lottery.java
kozhukhovsky: nor fixed

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 101;
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
+
+    public static Ball getRandomBall() {
+        String color = colorSupplier.getRandomColor();
+        int randomNumber = new Random().nextInt(MAX_BALL_NUMBER);
```
kozhukhovsky, line 11: move Random to the class field

kozhukhovsky (approved):
Approve, because its last attempt. But need fix :)

PR #1426, opened 2024-03-19
initial solution

MarcusPy:
In ColorSupplier class I decided to make field colors a static final because in my research, I found out that it will be more performance friendly this way, considering we want to fetch values from the Enum only once throughout the entire lifetime of the program. I also made the method getRandomColor static because we use it as a utility, so creating an instance seems redundant here. I hope the choices I made are right

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,10 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball.toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,7 +3,7 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
-    public Color getRandomColor() {
+    public static Color getRandomColor() {
         Random random = new Random();
         Color[] colors = Color.values();
         int randomIndex = random.nextInt(colors.length);
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        Random random = new Random();
+        Color[] colors = Color.values();
+        int randomIndex = random.nextInt(colors.length);
+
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Color[] colors = Color.values();
+
+    public static Color getRandomColor() {
+        Random random = new Random();
```
nowickimj, line 9: Make this a constant.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Color color = ColorSupplier.getRandomColor();
+        Random random = new Random();
```
nowickimj, line 8: this can be a constant as well.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Color[] colors = Color.values();
+
+    public static Color getRandomColor() {
```
nowickimj, line 8: This method shouldn't be static

nowickimj (changes requested):

MarcusPy:
Addressed the issues You've highlighted, although is storing an instance of Random as a constant a good practice? I've read that it is more performance friendly, but also reduces the randomness of it due to everything being generated from a single seed

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println(ball.toString());
```
```suggestion
            System.out.println(lottery.getRandomBall());
```
toString is called under the hood on the argument of println

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+
+    public Ball getRandomBall() {
+        Color color = new ColorSupplier().getRandomColor();
+        int randomNumber = random.nextInt(101);
```
okuzan, line 10: 101 should be a class level constant

okuzan (changes requested):

nowickimj (approved):
I will approve this PR because this is the final iteration, but one small remark: when creating constants, please use UPPER_CASE for names.

Regarding using Random instance as a constant - that's very good catch, although we may focus on clean code practices for now. There are better ways for proper randomization :)

PR #1431, opened 2024-03-22
Created Lottery, enum Colors, Ball classes. Implement method getRando…

VadymOziPantielieienko:
…mColor() for class ColorSupplier. Created a ball method with a random color and number. Created 3 balls from main() method.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,23 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        // ves proces randoma proishodit v classe lotery poetomy sozdaem ego kopiy
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+        // berem shar y kotorogo mi yje est pola color i number,
+        // i cherez loteri zovel metod getRandomBall v kotorom proishodit ves random
+        printResult(ball1);
+        printResult(ball2);
+        printResult(ball3);
+        // zovem nije napisanij metod i vstavlyaem shto vivesti
+    }
+
+    public static void printResult(Ball ball) {
+        // pechat rezyltata. (Ball ball) potomy shto metod toString lejit v Ball.java
+        System.out.println(ball.toString());
+        // nam nado vivesti na ekran a kakim vidom napisano v
+        // classe Ball potomy ot tyda i (Ball ball)
     }
 }
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,29 @@
+package core.basesyntax;
+
+public class Ball {
+
+    private String color;
+    private int number;
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    @Override
+    public String toString() { // override to stringa
+        return "Random color is: " + color + "\n" + "Random number is: " + number; // pishem kak hotim
+        // videt vidod
+    }
+}
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public String getRandomColor() { // vozvrashaem String
+        int index = new Random().nextInt(Colors.values().length); // генерує випадкове
+        // ціле число від 0 (включно) до розміру enum (не включно).
+        // Це число представляє індекс випадкової константи.
+        Colors color = Colors.values()[index]; //отримує константу enum
+        // за випадково згенерованим індексом.
+        return String.valueOf(color); // delaem stringom
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
+    // enum class neizmennye zveta
+    BLUE,
+    RED,
+    GREEN,
+    BLACK,
+    PURPLE,
+    PINK,
+    ORANGE
+}
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,21 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        // Ball potomy shto bydel brat ball i ego vozvrashat i
+        // prisvaivat znacheniya kotorie tam inicializirovany
+        Ball ball = new Ball(); // sozdaem novii Ball
+        ColorSupplier colorSupplier = new ColorSupplier();
+        // sozdaem noviy konstryktor kotorii generit cvet
+        ball.setColor(colorSupplier.getRandomColor());
+        // prisvaivaem cvet k ball istpolzya setter v
+        // kotorom berem metod random color shtobi zarandomit color
+        Random random = new Random(); // zovem konstryktor
+        // kotoriy daet nam vozmojnost randomit chisla
+        ball.setNumber(random.nextInt(100)); //prisvaivaem chislo
+        // cherez setter i randomim ego s nextInt. 100 eto predel 0-100
+        return ball; // retun ego
+    }
+}


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,21 @@
...
+        // sozdaem noviy konstryktor kotorii generit cvet
+        ball.setColor(colorSupplier.getRandomColor());
+        // prisvaivaem cvet k ball istpolzya setter v
+        // kotorom berem metod random color shtobi zarandomit color
+        Random random = new Random(); // zovem konstryktor
+        // kotoriy daet nam vozmojnost randomit chisla
+        ball.setNumber(random.nextInt(100)); //prisvaivaem chislo
+        // cherez setter i randomim ego s nextInt. 100 eto predel 0-100
+        return ball; // retun ego
+    }
```
MentorMA, line 20: image

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public String getRandomColor() { // vozvrashaem String
+        int index = new Random().nextInt(Colors.values().length); // генерує випадкове
+        // ціле число від 0 (включно) до розміру enum (не включно).
+        // Це число представляє індекс випадкової константи.
+        Colors color = Colors.values()[index]; //отримує константу enum
+        // за випадково згенерованим індексом.
+        return String.valueOf(color); // delaem stringom
```
MentorMA, line 12:
```suggestion
    public String getRandomColor() {
        int index = new Random().nextInt(Colors.values().length);      
        Colors color = Colors.values()[index]; 
        return String.valueOf(color);
```

MentorMA: clear all comments in code

MentorMA (changes requested):
Before sending PR read check list https://github.com/mate-academy/jv-lottery/blob/master/checklist.md

and clear all comments

src/main/java/core/basesyntax/Lottery.java
VadymOziPantielieienko: Changes added to code. Thank you for your feedback.

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum should be in singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Lottery getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
okuzan, line 12: 100 should be a class level constant with appropriate access modifier

okuzan, line 12: better create a constructor for ball

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,28 @@
...
+        this.number = number;
+    }
+
+    public String getColor() {
+        return color;
+    }
+
+    @Override
+    public String toString() {
+        return "Random color is: " + color + "\n" + "Random number is: " + number;
```
okuzan, line 26:
```suggestion
        return "color is: " + color + ", " + " number is: " + number;
```
what if we want to use this class in another context?

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
...
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[NUMBER_OF_BALLS];
+        for (int i = 0; i < lottery.length; i++) {
+            lottery[i] = new Lottery().getRandomBall();
+            printResult(lottery[i]);
+        }
+    }
+
+    public static void printResult(Lottery lottery) {
+        System.out.println(lottery.toString());
```
okuzan, line 15: not a mistake but am not sure if there's any point in extracting such a trivial piece of logic

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,17 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[NUMBER_OF_BALLS];
+        for (int i = 0; i < lottery.length; i++) {
+            lottery[i] = new Lottery().getRandomBall();
```
okuzan, line 9: Right now you create 3 lottery instances, you only need one - move lottery creation out of the scope of the loop

okuzan (changes requested):

src/main/java/core/basesyntax/Colors.java
VadymOziPantielieienko: got it. Always singular. Fixed

src/main/java/core/basesyntax/Lottery.java
VadymOziPantielieienko: Changed to private static final int

VadymOziPantielieienko: vey goood tip. thx. fixed.

src/main/java/core/basesyntax/Ball.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: I couldn't find another way in my head how to print it out)

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,19 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+    private static final Lottery lottery1 = new Lottery();
```
MAMentorFX, line 5: Why do you need this object as a field? When you have created another object inside main() (which is the correct approach)

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,19 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[NUMBER_OF_BALLS];
+        for (int i = 0; i < lottery.length; i++) {
+            lottery[i] = lottery1.getRandomBall();
```
MAMentorFX, line 10: lottery[i] is an object which is an instance of Lorttery class, so you can call getRandomBall() on it, you don't need lottery1.

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,19 @@
...
+            printResult(lottery[i]);
```
MAMentorFX, line 11: The same way here. Call toString() on lottery[i]

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,33 @@
+package core.basesyntax;
+
+public class Ball extends Lottery {
+
```
MAMentorFX, line 4:
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,19 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_BALLS = 3;
+    private static final Lottery lottery1 = new Lottery();
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lottery = new Lottery[NUMBER_OF_BALLS];
```
MAMentorFX, line 8: An array of lottery may contain only lottery objects. I guess, you thought, that this array would contain balls. For this task, there's no reason to create more than one lottery.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,33 @@
...
+        this.color = color;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    public String getColor() {
+        return color;
+    }
```
MAMentorFX, line 27: You don't really need getters and setters ffor this task

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
MAMentorFX, line 7: the problem with such approach, is that whenever you call this method, the object Random is created. Creating and maintaining object is some work that JVM has to do, but in your case there no need to create multiple Random object. Think, how you can create Random object only once for this ColorSupplier class.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+
```
MAMentorFX, line 6:
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+    private static final int maxNumber = 100;
```
MAMentorFX, line 7: Please, remember, how we name constants in Java.

MAMentorFX, line 7: also nextInt() returns int from 0 to `<` passed variable, not 0 `<= `

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,19 @@
...
+        for (int i = 0; i < lottery.length; i++) {
+            lottery[i] = lottery1.getRandomBall();
+            printResult(lottery[i]);
+        }
+
```
MAMentorFX, line 13:
```suggestion
```

MAMentorFX (changes requested):
Nice work, but needs extra fixes. This task is simpler than you think

src/main/java/core/basesyntax/Ball.java
VadymOziPantielieienko: when i did mvn clean package one of the mistakes was to create getters and setters that's why they are here

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: when i do this way i have an arr "lottery[I] is null" 
my code is  lottery[i] = lottery[I].getRandomBall();

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: im lost. we need 3 balls. to create 1 ball we need to call one lottery, am I r?

src/main/java/core/basesyntax/ColorSupplier.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Lottery.java
VadymOziPantielieienko: Fixed

src/main/java/core/basesyntax/Lottery.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: did it the same way I used it in the previous exercise

src/main/java/core/basesyntax/Ball.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: word object helps me. fixed

src/main/java/core/basesyntax/Lottery.java
VadymOziPantielieienko: fixed

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+
```
MAMentorFX, line 4:
```suggestion
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lotterys = new Lottery[]{
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall(),
+                new Lottery().getRandomBall()
+        };
```
MAMentorFX, line 10: That's not wrong. But not efficient. The same lottery may produce multiple Balls. So we don't need to create Lottery object for each ball. 
Imagine you have card deck and you want to play poker with your friends. With such approach, you don't want to reuse the same deck for the second game or the third one. You said to your friends "No, this deck is only for one game. Let's go buy another deck for the second game".

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery[] lotterys = new Lottery[]{
```
MAMentorFX, line 6:
```suggestion
        Lottery[] lotteries = new Lottery[]{
```

MAMentorFX (approved):
Well done. Just reuse one Lottery, please :)

src/main/java/core/basesyntax/Application.java
VadymOziPantielieienko: I understand what u mean but i guess i do not understand how to create this code so it will work that way. Would be nice if u show me an example that I can use later in my projects

MAMentorFX: > I understand what u mean but i guess i do not understand how to create this code so it will work that way. Would be nice if u show me an example that I can use later in my projects @MAMentorFX

Just create an object of Lottery, put it in variable. Create for-loop which iterates 3 times. Inside your loop call sout(lottery.getRandomBall().toString())

VadymOziPantielieienko: omg i always think about more complicated stuff

VadymOziPantielieienko: THX!!

PR #1432, opened 2024-03-24
add solution

maxkrup98:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
```
Vasul-Pankiv, line 6: Make Random object and colors arr a class field

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier color = new ColorSupplier();
+        Random random = new Random();
```
Vasul-Pankiv, line 8: same here, it is not a good idea to create this objects everytime you call method

Vasul-Pankiv (changes requested):

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+public enum Colors {
```
MAMentorFX, line 3:
```suggestion
public enum Color {
```
enum names are in singular

MAMentorFX (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
MAMentorFX: Not fixed. Yoyu don't want to create new Random each time you call getRandomColor(), you can reuse already created Random, which is declared on class level

MAMentorFX (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private ColorSupplier color = new ColorSupplier();
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(color.getRandomColor(),random.nextInt(1, 101));
```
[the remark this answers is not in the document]
MAMentorFX: Learn more about magic numbers. We move the to the constants, to clarify what those `1` and `101` means

MAMentorFX (approved):

PR #1438, opened 2024-03-30
added solution of the task

AndriiLoto:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier color = new ColorSupplier();
+
+    public String getRandomBall() {
+        int number = new Random().nextInt(100);
```
okuzan, line 9: 100 -> class-level constant
Random -> class-level field
don't forget appropriate access modifiers, constant case

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier color = new ColorSupplier();
+
+    public String getRandomBall() {
```
okuzan, line 8: why do you limit yourself with this return type? You should return a regular Ball object, it's up to the client/user to transform it to the String if it's needed, but if someone needed a whole object, he would need to parse it from String :)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+        int number = new Random().nextInt(100);
+        return new Ball(color.getRandomColor(),Integer.valueOf(number).toString()).toString();
```
okuzan, line 10: use IDEA's auto formatter (Ctrl Alt L), there should be a space after a comma
```suggestion
        return new Ball(color.getRandomColor(), number);
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
enum should be in singular

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
okuzan, line 7: same for this random, no need to recreate it per each method invocation

okuzan (changes requested):
Good job on your first try!

MentorMA (approved):

PR #1440, opened 2024-04-02
Сreate a simple lottery that return balls with a random color and ran…

SviatoslavTataryn:
…dom number.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -7,6 +7,6 @@ public class ColorSupplier {
 
     public String getRandomColor() {
         int index = random.nextInt(Colors.values().length);
-        return Colors.values()[index].toString();
+        return Colors.values()[index].name();
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,7 +3,7 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
-    private static final Random random = new Random();
+    private final Random random = new Random();
 
     public String getRandomColor() {
         int index = random.nextInt(Colors.values().length);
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private static final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Colors.values().length);
+        return Colors.values()[index].toString();
```
kerrrusha, line 10: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants
```suggestion
        return Colors.values()[index].name();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+    private static final ColorSupplier colorSupplier = new ColorSupplier();
```
kerrrusha, line 7: its fields not constants, so they should not be static
```suggestion
    private final Random random = new Random();
    private final ColorSupplier colorSupplier = new ColorSupplier();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+
+    public Ball getRandomBall() {
+        int number = random.nextInt(100) + 1;
```
kerrrusha, line 10: 100 is magic number

kerrrusha (changes requested):
overall is nice

src/main/java/core/basesyntax/Lottery.java
kerrrusha: not fixed

src/main/java/core/basesyntax/Lottery.java
kerrrusha: not fixed

kerrrusha (changes requested):

PR #1455, opened 2024-04-12
buch of code added

Kanril:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import static core.basesyntax.Lottery.getRandomBall;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        getRandomBall();
+        getRandomBall();
+        getRandomBall();
```
nowickimj, line 9: If you have to perform an action multiple times, better use a loop.

https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-for-loop-for-creating-several-objects-of-the-same-class

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball extends ColorSupplier {
+    private String color;
+    private int number;
+
+    public Ball() {
+        this.number = new Random().nextInt(101);
```
nowickimj, line 10: `Random` object should be a field. Or you can use static method `Math.random()` without declaring any new object.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
...
+        this.color = super.getRandomColor();
+    }
+
+    public String getBall() {
+        StringBuilder builder = new StringBuilder();
+        builder.append(number);
+        String out = color + " " + builder.toString();
+        return out;
```
nowickimj, line 18: You don't need a string builder here, just return the result of concatenation:

https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#dont-create-temporary-variables-if-you-can-directly-return-value-from-the-method

```java
return "color: " + color + ...;
```

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Ball extends ColorSupplier {
```
nowickimj, line 5: `Ball` is not meant to be an extension for `ColorSupplier` - those are two separate entities in the context of this application.
Ball IS NOT a color supplier.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,20 @@
...
+    private String color;
+    private int number;
+
+    public Ball() {
```
nowickimj, line 9: Don't generate values inside the constructor - you should pass them as constructor arguments

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
nowickimj, line 7: Same here - either make `Random` a constant or use `Math.random()`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    public static void getRandomBall() {
+        Ball ball = new Ball();
+        System.out.println(ball.getBall());
```
nowickimj, line 7: Please pay attention to what happens here:
1. You create an instance of an ball.
2. You try to get ball... out of the ball.

What should happen here is:
1. Prepare random number and color
2. Create and return an instance of a ball with prepared values (passed to constructor).

nowickimj (changes requested):

Kanril:
done

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import static core.basesyntax.Lottery.getRandomBall;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
+            Ball ball =
+                    getRandomBall(new ColorSupplier().getRandomColor(),
```
nowickimj, line 9: You are creating a new `ColorSupplier` instance on each loop iteration - that's not good.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static Random random = new Random();
+
+    public static Random getRandom() {
```
nowickimj, line 8: This class is a color supplier - it shouldn't provide `Random` instance as it violates Single Responsibility Principle.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    public static Ball getRandomBall(Colors color, int number) {
```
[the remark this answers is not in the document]
Kanril: image
Fixed it, so now getRandomBall gets color and number inside method, not as input data

src/main/java/core/basesyntax/Application.java
Kanril: image
Fixed, now ColorSupplier object creating inside class , but outside the method , so just one ColorSupplier creates

src/main/java/core/basesyntax/ColorSupplier.java
Kanril: image
created new class called Random (may be it could be better to use Java API class , but i think that this cleaner for people to understand code)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
+        return random;
+    }
+
+    public Colors getRandomColor() {
+        Colors color = Colors.values()[random.nextInt(Colors.values().length)];
+        return color;
```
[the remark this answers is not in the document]
Kanril: image
Fixed

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 
+import static core.basesyntax.Lottery.getRandomBall;
+
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        for (int i = 0; i < 3; i++) {
+            Ball ball =
+                    getRandomBall();
```
nowickimj, line 10: Unnecessary line break after `=`

src/main/java/core/basesyntax/Random.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public class Random {
+    private static java.util.Random random = new java.util.Random();
```
nowickimj, line 4: As much as I like the idea, you're not supposed to use static methods in this exercise: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#dont-use-static-methods-in-your-solution

Proper util class would require you to implement the singleton pattern, which might be a bit challenging now, so let's stick with non-static objects.
Besides that, there's already a static way to get random numbers: `java.lang.Math.random()`

Please remove this class.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    private static ColorSupplier colorSupplier = new ColorSupplier();
```
nowickimj, line 5: Don't make it static.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+    public static Ball getRandomBall() {
```
nowickimj, line 7: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#dont-use-static-methods-in-your-solution

nowickimj (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Lottery {
+
+    private ColorSupplier colorSupplier = new ColorSupplier();
+
+    public Ball getRandomBall() {
+
+        Colors color = colorSupplier.getRandomColor();
+        int number = (int) (Math.random() * 100) + 1;
```
nowickimj, line 10: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants

nowickimj (changes requested):
Looks good, one more thing to fix.

nowickimj (approved):

PR #1485, opened 2024-05-03
homeworkdone

dkisly:


src/main/java/core/basesyntax/Color.java
```diff
@@ -0,0 +1,6 @@
+package core.basesyntax;
+
+enum Color {
+        red, green, blue, yellow, orange;
+}
```
MentorMA, line 5: Check how you need to setup enum

MentorMA (changes requested):
before sending task read checklist

PR #1486, opened 2024-05-04
Hw jv lottery solving the tasks 

snowfallen:


okuzan (approved):
Well done on the first try!

PR #1491, opened 2024-05-09
added an implementation for the application Lottery

mstefanchuk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
```
MentorMA, line 8: return color in String type

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int maxRandomNumber = 100;
```
MentorMA, line 6: It should be constant

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public class Ball {
+    private Color color;
```
MentorMA, line 4:
```suggestion
    private String color;
```

MentorMA (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        Color[] colors = Color.values();
+        int randomIndex = random.nextInt(colors.length);
+        return String.valueOf(colors[randomIndex]);
```
sarakhmen, line 11: it's better to use name() method here

sarakhmen (changes requested):
Let's improve your solution following common mistakes :)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i <= 3; i++) {
+            Ball ball = lottery.getRandomBall();
+            System.out.println("ball " + i + ": " + ball);
```
okuzan, line 8: In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass/use the expression directly without assigning it to a variable

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 1; i <= 3; i++) {
```
okuzan, line 6: not a mistake, but get used to start iterating from 0

okuzan (approved):
Good job!

PR #1494, opened 2024-05-14
added solution of task according to README.md

apanyutin:


okuzan (changes requested):
Please, fix all mistakes mentioned here and then we will be able to review 💖

apanyutin:
I fixed some things tied with "static". Do I need to do a new pull request or you will review this?

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
```
Elena-Bruyako, line 6: This method should create a ball with a random color and a random number (the maximum possible number should be 100) and return this ball from the method.
What in Java can we use to iterate a part of program several times and avoid creating Ball line by line?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball(Color color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public Color getColor() {
```
Elena-Bruyako, line 12: Remove unused methods

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        Random random = new Random();
```
Elena-Bruyako, line 7: [Think about which variables should be local in the method and which should be class-level](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

src/main/java/core/basesyntax/Application.java
Elena-Bruyako: Hint: https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-for-loop-for-creating-several-objects-of-the-same-class

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        int randomNumber = random.nextInt(Color.values().length);
+        return Color.values()[randomNumber];
```
Elena-Bruyako, line 9: [Use name() for getting String representation of enum constants](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        int randomNumber = new Random().nextInt(100);
```
Elena-Bruyako, line 7: [All magic numbers in your code should be constants.](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,11 @@
...
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Elena-Bruyako, line 8: [Think about which variables should be local in the method and which should be class-level](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
```
Elena-Bruyako, line 6:
```suggestion
    public String getRandomColor() {
```

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Ball.java
apanyutin: Why do we need to do this? We could use them in the future.

src/main/java/core/basesyntax/ColorSupplier.java
apanyutin: I don't understand what you mean, in this part of code method getRandomColor() returns type Color is not String.

src/main/java/core/basesyntax/ColorSupplier.java
apanyutin: Why we need to do this? Can the return type be a Enum type? Class Ball has field Color of Enum type.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,15 @@
...
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (int i = 0; i < balls.length; i++) {
+            System.out.println("Ball №" + (i + 1) + " - " + balls[i]);
+        }
```
kerrrusha, line 14: should be combined into single loop

kerrrusha (approved):

PR #1501, opened 2024-05-25
homework 3 try 1

vdsfww:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
+        System.out.println(lottery.getRandomBall());
```
sarakhmen, line 8: let's do it in a loop instead

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "color - " + color + ", number - " + number;
+    }
```
sarakhmen, line 15: I'd recommend generating toString method using Intellij Idea

sarakhmen (changes requested):
Good job, but let's slightly improve your solution :)

src/main/java/core/basesyntax/Application.java
vdsfww: it`s look like the
for (int i = 0; i < 3; i++) {
}

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako, line 6: Make 3 as a constant

Elena-Bruyako (approved):

PR #1508, opened 2024-05-30
JV-LOTTERY task done

APekariev:
JV-LOTTERY task complete

### DIFF
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -5,7 +5,6 @@ import java.util.Random;
 public class Lottery {
     private static final int RANGE_BOUND = 100;
     private ColorSupplier colorSupplier;
-    private Ball ball = new Ball();
     private Random random = new Random();
 
     public Lottery(ColorSupplier colorSupplier) {
@@ -13,6 +12,7 @@ public class Lottery {
     }
 
     public Ball getRandomBall() {
+        Ball ball = new Ball();
         ball.setColor(colorSupplier.getRandomColor());
         ball.setNumber(random.nextInt(RANGE_BOUND));
         return ball;


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,8 @@
+package core.basesyntax;
+
+public enum Colors {
```
kerrrusha, line 3: enum and class names should be in singular not plural
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+public class Lottery {
+    private ColorSupplier colorSupplier;
+
+    public Lottery(ColorSupplier colorSupplier) {
+        this.colorSupplier = colorSupplier;
+    }
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        Random random = new Random();
```
kerrrusha, line 14: should be class-level field, like `colorSupplier`

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,19 @@
...
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(100));
```
kerrrusha, line 16: 100 is a magic number

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        Colors color = Colors.values()[index];
+        return color.toString();
```
kerrrusha, line 9: use name instead of toString, refer to the checklist

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
kerrrusha, line 7: random should be class-level field

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,27 @@
...
+        return this.color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return this.number;
+    }
```
kerrrusha, line 17: never used
```suggestion
    public void setColor(String color) {
        this.color = color;
    }
```

kerrrusha (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int RANGE_BOUND = 100;
+    private ColorSupplier colorSupplier;
+    private Ball ball = new Ball();
```
okuzan, line 8: this should be a local variable - we create a whole new object each time this method is called, its scope should be local to the method. It's not the same as ColorSupplier - we don't need to recreate it, only use it

okuzan (changes requested):
Almost there!

sarakhmen (approved):
Good job!

PR #1525, opened 2024-06-21
Hw 02 jv lottery

skrypkasv:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako, line 6: 3 is a magic number, create constant

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
```
Elena-Bruyako, line 7: Check which get and set methods you actually use and remove unused

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
+import core.basesyntax.enums.Color;
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
+    public Random getRandom() {
```
Elena-Bruyako, line 9: remove

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,16 @@
...
+        return random;
+    }
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[random.nextInt(Color.values().length)].name();
```
Elena-Bruyako, line 14: It's hard to read such code, it's better to create variable `index = random.nextInt(Color.values().length)`  here

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
skrypkasv: only getRandom()? or both: field random + getRandom()???

src/main/java/core/basesyntax/ColorSupplier.java
skrypkasv: In my third commit I did exactly the same var, but after reading COMMON MISTAKES I've removed it. I agree with your point of view and I'll fix it.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class Lottery {
+    private static final int NUMBER_BOUNDARY = 101;
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(random.nextInt(NUMBER_BOUNDARY));
```
okuzan, line 13: create a constructor for those two fields

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+        return ball.toString();
```
okuzan, line 14: return Ball itself, not string representation. This is more flexible. Maybe some day you will need ball itself, not just string. If you need string you can call toString in a caller method (though it's not needed in System.out.println - it's called under the hood)

okuzan (changes requested):

src/main/java/core/basesyntax/Lottery.java
skrypkasv: +

src/main/java/core/basesyntax/Lottery.java
skrypkasv: Could you explain me exactly what do you mean 'those two fields'???

src/main/java/core/basesyntax/Lottery.java
skrypkasv: ok

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        final int ballsLimit = 3;
```
Elena-Bruyako, line 5: it should be constant and placed as class-level
`private static final int NUMBER_OF_BALLS = 3; `

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import core.basesyntax.enums.Color;
+import java.util.Random;
+
 public class ColorSupplier {
+    private final int index = new Random().nextInt(Color.values().length);
```
Elena-Bruyako, line 7:
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
...
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[index].name();
```
Elena-Bruyako, line 10:
```suggestion
        int index = random.nextInt(Color.values().length);
        return Color.values()[index].name();
```

src/main/java/core/basesyntax/Lottery.java
Elena-Bruyako: Probably mentor suggested to create constructor in Ball class. In this case, you can remove set methods in class Ball
`public Ball(String color, int number) {}`

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Lottery.java
skrypkasv: Now it's understandable. Thanks

src/main/java/core/basesyntax/Application.java
skrypkasv: Are you sure that we can use 'privat static' modifiers in this case??? 
It will lead to an error 'illegal start of expression'

skrypkasv: ok

src/main/java/core/basesyntax/ColorSupplier.java
Elena-Bruyako: Please, follow the comments. If you have some questions, please ask. 

You should create class-leve variable Random, to avoid creating it each time, when you call getRandomColor() method. 

I left this comment 
`private final Random random = new Random(); `

And in method getRandomColor() you just create `int index = random.nextInt(Color.values().length); `

Elena-Bruyako (changes requested):
Be attentive to the comments, if you have some questions, ask in chat

Elena-Bruyako (approved):

PR #1539, opened 2024-07-06
Final example

Andormy:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,12 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Random random = new Random();
+
+    public Color getRandomColor() {
```
MentorMateAcademy, line 8:
```suggestion
    public String getRandomColor() {
```

MentorMateAcademy: return color in String format

MentorMateAcademy (changes requested):

MAMentorFX (approved):

PR #1540, opened 2024-07-07
Homework solution. Create simple lottery.

Extase67:


MentorMateAcademy (changes requested):

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,20 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 101;
```
MAMentorFX, line 6:
```suggestion
    private static final int MAX_NUMBER = 100;
```
do not change requirements, handle nextInt() with max number 100

MAMentorFX (approved):

PR #1541, opened 2024-07-07
complete jv-lottery

SerhiiStupak:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako, line 5: [All magic numbers in your code should be constants.](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants)

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
...
+            System.out.println(new Lottery().getRandomBall());
```
Elena-Bruyako, line 6: [Don't create several instances of the same class if you can use only one instance for your purposes](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
Elena-Bruyako, line 7: [Think about which variables should be local in the method and which should be class-level](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), new Random().nextInt(100));
```
Elena-Bruyako, line 7: [Think about which variables should be local in the method and which should be class-level](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#think-about-which-variables-should-be-local-in-the-method-and-which-should-be-class-level)

Elena-Bruyako, line 7: [All magic numbers in your code should be constants.](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants)

Elena-Bruyako (changes requested):

MentorMateAcademy (approved):

PR #1546, opened 2024-07-16
Home work lottery

4Vitalii5:
Create class Ball with fields (color, number) and override method toString.
Create enum Colors.
Implement method getRandomColor.
Create class Lottery with getRandomBall method.
Added logic to test the application.

### DIFF
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,0 +1,30 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
+
+    public String getColor() {
+        return color;
+    }
+
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    public int getNumber() {
+        return number;
+    }
+
+    public void setNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Ball{"
+                + "color='" + color + '\''
+                + ", number=" + number
+                + '}';
+    }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,30 @@
+package core.basesyntax;
```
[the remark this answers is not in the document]
4Vitalii5: > Please add an access modifier to your Ball class. It's a good practice to define the scope of your classes explicitly, even if you want to use the default package-private access level.

I have a access modifier public in this class.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        Ball ball = new Ball();
+        int randomNumber = random.nextInt(100);
```
MentorMateAcademy, line 11: magic number

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+        ball.setNumber(randomNumber);
+        ball.setColor(colorSupplier.getRandomColor());
```
MentorMateAcademy, line 13: Better create ball using constructor

MentorMateAcademy (changes requested):

MentorMateAcademy (approved):

PR #1547, opened 2024-07-17
Solution

misha-doroshenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public Color getRandomColor() {
+        int index = new Random().nextInt(Color.values().length);
+        return Color.values()[index];
     }
```
liliia-ponomarenko, line 9: In this case, you create an instance of Random each time you call this method. Declare instances at the class field level and initialize in the field or the constructor

```suggestion
    private final Random random;
    
    public ColorSuplier(Random random) {
      this.random = new Random();
    }
    public Color getRandomColor() {
        int index = random.nextInt(Color.values().length);
        return Color.values()[index];
    }
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,7 @@
+package core.basesyntax;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor());
```
liliia-ponomarenko, line 5: Let's initialize the ColorSupplier as I suggested above, also passes the number to the Ball constructor ;)

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.Random;
+
+public class Ball {
+    private Color color;
+    private int number;
+
+    public Ball(Color color) {
+        this.color = color;
+        this.number = new Random().nextInt(100);
```
liliia-ponomarenko, line 11: All magic numbers in your code should be constants.

liliia-ponomarenko (changes requested):

PR #1550, opened 2024-07-21
created three instances of the Ball class with random attributes

GarikSlavsky:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,12 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[3];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
+        }
```
Elena-Bruyako, line 11:
```suggestion
        Lottery lottery = new Lottery();

        for (int i = 0; i < USE_CONSTANT; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

Create constant for 3 magic number and use it

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,30 @@
...
+    private String color;
+    private int number;
+
+    public int getNumber() {
+        return number;
+    }
+
+    public String getColor() {
+        return color;
+    }
```
Elena-Bruyako, line 13:
```suggestion
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
Elena-Bruyako, line 3:
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUND = 101;
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Elena-Bruyako, line 10: ColorSupplier should be class-level variable

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Ball.java
GarikSlavsky: Why should the getters be deleted?

Elena-Bruyako (approved):

src/main/java/core/basesyntax/Lottery.java
GarikSlavsky: Why should the 'colorSupplier' be a class-level variable while the 'ball' is located within the method?

PR #1556, opened 2024-08-01
hw2 lottery

bilsinas:
created a method of random color and number generation. For the test, I used the creation of 3 balls


okuzan (changes requested):
Please check if it builds locally first with command like `mvn clean package`, only if build passes create a PR. if you have any troubles please contact mentors on the platform! We are here to help

PR #1559, opened 2024-08-02
corrected checkstyle

bilsinas:


okuzan (changes requested):
Please look through these mistakes first, I some of them in your code, and then we can continue with the review 🤝

PR #1560, opened 2024-08-03
Implemented Lottery

rainett:
Created and implemented Ball, Color and Lottery classes. Implemented ColorSupplier and Application classes

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -2,6 +2,9 @@ package core.basesyntax;
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako, line 6: 3 is magic number

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,32 @@
...
+public class Ball {
+    private String color;
+    private int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    public String getColor() {
```
Elena-Bruyako, line 12: check that you use all these get and set methods, if not - remove unused

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1571, opened 2024-08-12
Solution

vsupruniuk:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        Random random = new Random();
```
liliia-ponomarenko, line 7: Think about which variables should be local in the method and which should be class-level. 
Each time you call this method, a new random instance will be created. Let's make it a class variable and initialize it in the constructor

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,10 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```
use singular in class names

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        Random random = new Random();
```
liliia-ponomarenko, line 7: the same about Random

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+        String color = new ColorSupplier().getRandomColor();
```
liliia-ponomarenko, line 9: Make ColorSupplier class-level variable. See details in the common mistakes list;)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        int number = random.nextInt(100);
```
liliia-ponomarenko, line 10: All magic numbers in your code should be constants.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,14 @@
...
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        Ball ball1 = lottery.getRandomBall();
+        Ball ball2 = lottery.getRandomBall();
+        Ball ball3 = lottery.getRandomBall();
+
+        System.out.println(ball1);
+        System.out.println(ball2);
+        System.out.println(ball3);
```
liliia-ponomarenko, line 13: Use for loop for creating several objects of the same class (common mistakes)

liliia-ponomarenko (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,17 @@
...
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[ballsCapacity];
+
+        for (int i = 0; i < ballsCapacity; i += 1) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
```
liliia-ponomarenko, line 16: you can do it using one loop ;)

liliia-ponomarenko (approved):
Great job!

PR #1577, opened 2024-08-17
solution

oleksandrshtonda:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        System.out.println(lottery.getRandomBall());
```
Elena-Bruyako, line 7: use for loop

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final Colors color;
```
Elena-Bruyako, line 4:
```suggestion
    private final String color;
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private final Random random = new Random();
+    private final Colors[] colorArr = Colors.values();
+
+    public Colors getRandomColor() {
```
Elena-Bruyako, line 9:
```suggestion
    public String getRandomColor() {
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+public enum Colors {
```
Elena-Bruyako, line 3:
```suggestion
public enum Color {
```

read enum naming convention, it should be singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private final int maxRandomNumber = 100;
```
Elena-Bruyako, line 6:
```suggestion
    private static final int maxRandomNumber = 100;
```
make it upper-case

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 10; i++) {
```
Elena-Bruyako, line 7: You should create three balls using the getRandomBall() method from the Lottery class. 

Also, 10 is magic number, create constant

Elena-Bruyako (changes requested):

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+    private final Color[] colorArr = Color.values();
```
okuzan, line 7:
```suggestion
    private final Color[] colors = Color.values();
```

okuzan (approved):

PR #1578, opened 2024-08-18
apply solution

oksana-feshchenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,10 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
     public String getRandomColor() {
-        return null;
+        return Color.values()[new Random().nextInt(0, Color.values().length)].name();
```
d1sam, line 8: please make random object as a class member

d1sam (changes requested):
To test the program, in the main() method, create three balls using the getRandomBall() method from the Lottery class. And print information about them in the console.
fix all the issues described.

PR #1605, opened 2024-09-16
Update ColorSupplier.java

katrienkraska:
create an enum with different colors
implement method getRandomColor() in class ColorSupplier. This method should return a random color, based on the enum values (use the Random class for this)


Elena-Bruyako (changes requested):
there is only one class in PR, provide full implementation

PR #1632, opened 2024-09-19
Fisrst approach to Lottery task.

Pawel-Cp:
Some fields should be final. Could you tell me why?

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -7,8 +7,12 @@ public class Application {
 
         for (int i = 0; i < 3; i++) {
             Ball ball = lottery.getRandomBall();
+            if (ball == null) {
+                throw new IllegalArgumentException("Invlaid value");
+            }
             System.out.println(ball);
         }
     }
 }
+
 
--- a/src/main/java/core/basesyntax/Ball.java
+++ b/src/main/java/core/basesyntax/Ball.java
@@ -1,16 +1,16 @@
 package core.basesyntax;
 
 public class Ball {
-    private final String color;
+    private final Color color;
     private final int number;
 
-    public Ball(String color, int number) {
+    public Ball(Color color, int number) {
         this.color = color;
         this.number = number;
     }
 
     @Override
     public String toString() {
-        return "The number is: " + number + " the color is: " + color;
+        return "Ball #" + number + " is " + color;
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -4,11 +4,14 @@ import java.util.Random;
 
 public class ColorSupplier {
 
-    private final Random randomColor = new Random();
+    private final Random random = new Random();
 
-    public String getRandomColor() {
-        Colors[] colors = Colors.values();
-        int randomIndex = randomColor.nextInt(colors.length);
-        return colors[randomIndex].name();
+    public Color getRandomColor() {
+        Color[] colors = Color.values();
+        if (colors == null || colors.length == 0) {
+            throw new IllegalArgumentException("Empty array");
+        }
+        int randomIndex = random.nextInt(colors.length);
+        return colors[randomIndex];
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -3,7 +3,7 @@ package core.basesyntax;
 import java.util.Random;
 
 public class Lottery {
-    private static final int MAX_VALUE = 101;
+    private static final int MAX_VALUE = 100;
 
     private final ColorSupplier colorSupplier = new ColorSupplier();
     private final Random randomValueSupplier = new Random();
--- a/src/main/java/core/basesyntax/Color.java
+++ b/src/main/java/core/basesyntax/Color.java
@@ -1,6 +1,6 @@
 package core.basesyntax;
 
-public enum Colors {
+public enum Color {
     BLACK,
     WHITE,
     RED,


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
slade13, line 3: The enum name 'Colors' should be singular 'Color' as each instance of an enum represents a single value.

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 101;
```
slade13, line 6: The value of MAX_VALUE should be 100, not 101. The task specifies that the maximum number for a ball should be 100. You can add 1 to this value in random number generation, but this constant should be 100.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int number;
+
+    public Ball(String color, int number) {
+        this.color = color;
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "The number is: " + number + " the color is: " + color;
```
slade13, line 14: The toString method is not formatted properly. It should return a string formatted as 'ball #{number} is {color}', not 'The number is: {number} the color is: {color}'.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
+package core.basesyntax;
+
+public class Ball {
+    private final String color;
```
slade13, line 4: The task specified to define an enum for different colors. You should not use String for color field in Ball class, but a type of the defined enum.

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,16 @@
...
+    private final int number;
+
+    public Ball(String color, int number) {
```
slade13, line 7: The constructor should take an argument of the color enum type, not a String.

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            Ball ball = lottery.getRandomBall();
```
slade13, line 9: The 'getRandomBall()' function is called without checking if it returns a valid object. It's a good practice to ensure that the function does not return 'null' before using its returned object.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
+
 public class ColorSupplier {
+
+    private final Random randomColor = new Random();
+
     public String getRandomColor() {
-        return null;
+        Colors[] colors = Colors.values();
+        int randomIndex = randomColor.nextInt(colors.length);
+        return colors[randomIndex].name();
```
slade13, line 12: The getRandomColor() method returns a string instead of the enum value. Consider changing the method return type from String to Colors.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
+
+    private final Random randomColor = new Random();
```
slade13, line 7: The variable 'randomColor' is misleading. It's not a color, it's a Random object. Consider renaming it to 'random'.

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,14 @@
...
+
     public String getRandomColor() {
```
slade13, line 9: Consider adding a null check for an empty array of colors in the getRandomColor() method. If you attempt to access an element of an empty array, it will cause an ArrayIndexOutOfBoundsException.

slade13 (changes requested):
Hey, you will need to fix few things before we proceed. Good luck!

slade13:
> Some fields should be final. Could you tell me why?

If value or reference mustn't be changed we can ensure such behavior with `final` keyword. https://brytecode.be/articles/use-of-the-final-keyword-in-java/#final-fields

src/main/java/core/basesyntax/Colors.java
Pawel-Cp: done

src/main/java/core/basesyntax/Lottery.java
Pawel-Cp: done

src/main/java/core/basesyntax/Ball.java
Pawel-Cp: done

src/main/java/core/basesyntax/Ball.java
Pawel-Cp: done

src/main/java/core/basesyntax/Ball.java
Pawel-Cp: done

src/main/java/core/basesyntax/Application.java
Pawel-Cp: done

src/main/java/core/basesyntax/ColorSupplier.java
Pawel-Cp: done

src/main/java/core/basesyntax/ColorSupplier.java
Pawel-Cp: done

src/main/java/core/basesyntax/ColorSupplier.java
Pawel-Cp: I think I've done this correctly but the Inteliij idicate warning about "colors == null is always false.
Should I do it differently?

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_VALUE = 100;
+
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random randomValueSupplier = new Random();
+
+    public Ball getRandomBall() {
+        int number = randomValueSupplier.nextInt(MAX_VALUE);
```
slade13, line 12: You've changed MAX_VALUE correctly, but now in nextInt() you have add 1, so it will randomize a number between 0 and 100 inclusively. Here are the docs: https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#nextInt-int-
Your code should look like:
`int number = randomValueSupplier.nextInt(MAX_VALUE + 1);`

slade13 (approved):
Good job, just remember to change range in Lottery.java

PR #1633, opened 2024-09-20
complited the assigment

gubber230:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,13 @@
...
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball firstBall = lottery.getRandomBall();
+        Ball secondBall = lottery.getRandomBall();
+        Ball thirdBall = lottery.getRandomBall();
+
+        System.out.println(firstBall);
+        System.out.println(secondBall);
+        System.out.println(thirdBall);
```
okuzan, line 12: checklist issue - use a loop. This approach is not extensible, what if you had to print 30 balls?

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,19 @@
...
+    CYAN,
+    MAGENTA,
+    VIOLET,
+    BLACK,
+    WHITE,
+    BROWN,
+    GRAY,
+    PINK
+}
+
```
okuzan, line 19:
```suggestion
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,19 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+    private static final int MAX_BALL_NUMBER = 100;
+
+    public Ball getRandomBall() {
+        int number = random.nextInt(MAX_BALL_NUMBER);
+        String color = new ColorSupplier().getRandomColor();
```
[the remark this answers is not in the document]
okuzan: Don't create several instances of the same class if you can use only one instance for your purposes. Here, you can create a single instance of ColorSupplier outside the getRandomBall method and use it to get random colors.

okuzan (changes requested):
Good job overall!

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,15 @@
...
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[] {
+            lottery.getRandomBall(),
+            lottery.getRandomBall(),
+            lottery.getRandomBall()
+        };
+
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
```
okuzan, line 14: rewrite this solution without using an array,
just call needed method once - in a loop

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,18 @@
...
+        Ball[] balls = new Ball[LIMIT_ARRAY];
+
+        for (int i = 0; i < LIMIT_ARRAY; i++) {
+            balls[i] = lottery.getRandomBall();
+        }
+
+        for (Ball ball : balls) {
+            System.out.println(ball);
+        }
     }
```
okuzan, line 17:
```suggestion
        for (int i = 0; i < LIMIT_ARRAY; i++) {
            System.out.println(lottery.getRandomBall());
        }
    }
```

okuzan (changes requested):
Almost done

Elena-Bruyako (approved):

PR #1662, opened 2024-10-08
created simple lottery

sannchous:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public String getRandomBall() {
+        Ball ball = new Ball();
+        Random random = new Random();
+        ColorSupplier color = new ColorSupplier();
```
okuzan, line 9: make those two private class-level fields, to avoid needless recreation per each method call

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        ball.setNumber(random.nextInt(100));
```
okuzan, line 10: 100 should be a constant (private, static final)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+        ball.setColor(color.getRandomColor());
+        return ball.toString();
```
okuzan, line 12: return ball itself, let consumer trasform it to string if needed, our solution should allow more options than just string representation

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,29 @@
+package core.basesyntax;
+
+public class Ball {
+    private String color;
+    private int number;
```
okuzan, line 5: let's create a constructor for those values

okuzan (changes requested):
Good job overall

PR #1663, opened 2024-10-08
Solution:

nikolasdymchenko7:
Solution:
- Created new classes: Ball and Lottery
- Created Enum: Colors
- Added implementation to the method: getRandomColor()
- Added method getRandomBall to the Lottery class
- Created three balls in the Application class and printed them

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_LOTTERIES = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery[] lotteries = new Lottery[NUMBER_OF_LOTTERIES];
+        for (int i = 0; i < NUMBER_OF_LOTTERIES; i++) {
+            lotteries[i] = new Lottery();
+            System.out.println(lotteries[i].getRandomBall());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/Colors.java
+++ b/src/main/java/core/basesyntax/Colors.java
@@ -1,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
+    BLACK,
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+public enum Colors {
```
okuzan, line 3: enum should be in singular

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BOUNDARY_RANDOM = 100;
+
+    public Ball getRandomBall() {
+        int generatedNumber = new Random().nextInt(BOUNDARY_RANDOM);
```
okuzan, line 9: create Random and ColorSupplier at class level

okuzan (changes requested):

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_LOTTERIES = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery[] lotteries = new Lottery[NUMBER_OF_LOTTERIES];
+        for (int i = 0; i < NUMBER_OF_LOTTERIES; i++) {
+            lotteries[i] = new Lottery();
```
okuzan, line 10: don't use array at all
you need only one lottery

okuzan (changes requested):

src/main/java/core/basesyntax/Lottery.java
nikolasdymchenko7: @okuzan Sorry, could you explain why you need to create instances of Random and ClassSupplier at the class level?

PR #1687, opened 2024-11-01
Hm jv lottery solution

MaxMeleshchenko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Color.values().length);
```
ahoienko, line 7: It is better to create Random object on class level, because in current implementation you will create Random each time when method getRandomColor

ahoienko: Same for Lottery  class

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
+    public Ball getRandomBall() {
+        Random random = new Random();
+        int value = random.nextInt(100);
```
ahoienko, line 8: Magic number, please create a constant

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery extends ColorSupplier {
```
ahoienko, line 5: There is not reason to extend lottery from ColorSupplier, they are 2 totally different classes with different purpose, so they should not be in one hierarchy. So I suggest you to just create and object of ColorSupplier in Lottery and use getRandomColor method

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,9 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < 3; i++) {
```
ahoienko, line 6: 3 is also magic number, please move it to constant

ahoienko (changes requested):
Very nice, left few comments, please check)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
...
+import java.util.Random;
+
 public class ColorSupplier {
+    private final Random random = new Random();
+
     public String getRandomColor() {
-        return null;
+        int index = random.nextInt(Color.values().length);
+        Color color = Color.values()[index];
+        return color.toString();
```
Elena-Bruyako, line 11: [Use name() for getting String representation of enum constants](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_BALL_NUMBER = 100;
+    private final Random random = new Random();
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Ball ball = new Ball();
```
Elena-Bruyako, line 9:
```suggestion
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public Ball getRandomBall() {
+        int value = random.nextInt(MAX_BALL_NUMBER);
+        ball.setColor(colorSupplier.getRandomColor());
+        ball.setNumber(value);
```
Elena-Bruyako, line 14:
```suggestion
        Ball ball = new Ball;
        ball.setColor(colorSupplier.getRandomColor());
        ball.setNumber(random.nextInt(MAX_BALL_NUMBER));
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
 public class Application {
+    public static final int TOTAL_BALLS = 3;
```
Elena-Bruyako, line 4:
```suggestion
    private static final int TOTAL_BALLS = 3;
```

Elena-Bruyako (approved):

PR #1699, opened 2024-11-14
task has been completed

kimanko:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
```
liliia-ponomarenko, line 6: Don't use static methods in your solution

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
...
+        return Color.values()[new Random().nextInt(6)];
```
liliia-ponomarenko, line 7: Think about which variables should be local in the method and which should be class-level
Let’s  make 6 - the constant variable 
new Random() - class level variable

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        return new Ball(new Random().nextInt(101), ColorSupplier.getRandomColor());
+    }
```
liliia-ponomarenko, line 8: the same comments

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,8 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        System.out.println(Lottery.getRandomBall());
+        System.out.println(Lottery.getRandomBall());
+        System.out.println(Lottery.getRandomBall());
```
liliia-ponomarenko, line 7: Use for loop for creating several objects of the same class

liliia-ponomarenko (changes requested):
Let’s check the common mistakes list ;)

src/main/java/core/basesyntax/ColorSupplier.java
kimanko: I will change it but I thought its better to use static method since we do not need to create a new specific object of ColorSupplier as we this method will be same for any of those so we can just call it from our class. please explain me why I am wrong

kimanko: done

PR #1703, opened 2024-11-19
Implement Lottery

MaybeTI:


src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public enum Colors {
```
```suggestion
public enum Color {
```

liliia-ponomarenko (changes requested):
Let's check the common mistakes list and fix your solution by list ;)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    public static Color getRandomColor() {
```
liliia-ponomarenko, line 6:
```suggestion
    public Color getRandomColor() {
```
Don't use static methods in your solution (from common mistakes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,9 @@
...
+        return Color.values()[new Random().nextInt(Color.values().length)];
```
liliia-ponomarenko, line 7: Think about which variables should be local in the method and which should be class-level (following common mistakes)

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public static Ball getRandomBall() {
+        return new Ball(ColorSupplier.getRandomColor(), new Random().nextInt(101));
+    }
```
liliia-ponomarenko, line 8: the same Also, All magic numbers in your code should be constants. (from common mistakes)

liliia-ponomarenko (changes requested):
Let’s  check the common mistakes list again

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
 public class Application {
-    public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+    private static final int BALL_COUNT = 3;
+
+    public void main(String[] args) {
+        for (int i = 0; i < BALL_COUNT; i++) {
+            System.out.println(new Lottery().getRandomBall());
```
liliia-ponomarenko, line 8: in this realization, you create 3 new instances one instant is enough

liliia-ponomarenko, line 8:
```suggestion
    public void main(String[] args) {
        Lottery lottery = new Lottery();
        for (int i = 0; i < BALL_COUNT; i++) {
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private Random random = new Random();
```
liliia-ponomarenko, line 6:
```suggestion
    private final Random random = new Random();
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int BALL_NUMBER = 101;
+    private Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(new ColorSupplier().getRandomColor(), random.nextInt(BALL_NUMBER));
```
liliia-ponomarenko, line 10:
```suggestion
    private final Random random = new Random();
    private final ColorSupplier supplier = new ColorSupplier()

    public Ball getRandomBall() {
        return new Ball(supplier.getRandomColor(), random.nextInt(BALL_NUMBER));
```

liliia-ponomarenko (approved):
Good job! But fix my comments ;)

PR #1708, opened 2024-12-01
Create homework

PVladster:


homework
```diff
@@ -0,0 +1,17 @@
+package core.basesyntax;
+
```
Elena-Bruyako, line 2: you need to create a Ball class with the following fields: color and number and override the toString() method
create Lottery class with method
To test the program, in the main() method, create three balls using the getRandomBall() method from the Lottery class. And print information about them in the console.

Elena-Bruyako (changes requested):
before sending PR, check [common mistakes file ](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md)

PR #1834, opened 2025-05-07
feat: implement lottery ball generation

Lokkeez:
Random colors (from a predefined Color enum)
Random numbers (1-100)

Changes:

Added Ball class with color/number fields and toString().
Implemented ColorSupplier to pick random colors using java.util.Random.
Created Lottery class with getRandomBall() method.
Added basic tests in Main.

### DIFF
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Random random = new Random();
+
+    public static String getRandomColor() {
+        Color[] colors = Color.values();
+        int randomIndex = random.nextInt(colors.length);
+        return colors[randomIndex].name();
     }
 }
--- a/src/main/java/core/basesyntax/Lottery.java
+++ b/src/main/java/core/basesyntax/Lottery.java
@@ -1,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+
+    public static Ball getRandomBall() {
+        String color = ColorSupplier.getRandomColor();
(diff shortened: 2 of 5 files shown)


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,13 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    private static final Random random = new Random();
+
+    public static String getRandomColor() {
```
Elena-Bruyako, line 8:
```suggestion
    public String getRandomColor() {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final Random random = new Random();
+
+    public static Ball getRandomBall() {
```
Elena-Bruyako, line 8:
```suggestion
    public Ball getRandomBall() {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,13 @@
...
+        String color = ColorSupplier.getRandomColor();
+        int number = random.nextInt(100) + 1;
```
Elena-Bruyako, line 10: 100 is a magic number,  create constant

src/main/java/core/basesyntax/Main.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+public class Main {
+    public static void main(String[] args) {
+
+        Ball ball1 = Lottery.getRandomBall();
```
Elena-Bruyako, line 6: don't use numbers for naming
what if you need to create 50 instance of balls? let's use for loop

Elena-Bruyako (changes requested):

PR #1859, opened 2025-06-17
Feature/refactor lotteryRefactor: removed magic numbers, simplified main loop, made Random and ColorSupplier non-static

MaksymMaksymenko2:


MaksymMaksymenko2:
Implemented a lottery with the following classes: Ball, Color, ColorSupplier, and Lottery.

Removed static method from ColorSupplier; now it uses instance methods

Replaced magic numbers with constants

Added a loop in Main to create and print 3 balls

Fixed code style issues according to Checkstyle

PR #1874, opened 2025-07-14
jv-lottery-solution-Artem-Nurmetov

anurmetov:
added Classes: Ball (logic of creating ball object), ENUM Colors with some colors, ColorSupplier(logic of getting random color), main method to hardcode and test result.

### DIFF
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -3,5 +3,11 @@ package core.basesyntax;
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball firstBall = new Lottery().getRandomBall();
+        Ball secondBall = new Lottery().getRandomBall();
+        Ball thirdBall = new Lottery().getRandomBall();
+        System.out.println(firstBall);
+        System.out.println(secondBall);
+        System.out.println(thirdBall);
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,14 +1,13 @@
 package core.basesyntax;
 
 public class Application {
-    private static final int NUMBER_OF_USERS = 3;
+    private static final int NUMBER_OF_BALLS = 3;
 
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
-        Ball[] balls = new Ball[NUMBER_OF_USERS];
-        for (int i = 0; i < balls.length; i++) {
-            balls[i] = new Lottery().getRandomBall();
-            System.out.println(balls[i].toString());
+        Lottery lottery = new Lottery();
+        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
+            System.out.println(lottery.getRandomBall());
         }
     }
 }
--- a/src/main/java/core/basesyntax/Application.java
+++ b/src/main/java/core/basesyntax/Application.java
@@ -1,13 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
-        Ball firstBall = new Lottery().getRandomBall();
-        Ball secondBall = new Lottery().getRandomBall();
-        Ball thirdBall = new Lottery().getRandomBall();
-        System.out.println(firstBall);
-        System.out.println(secondBall);
-        System.out.println(thirdBall);
+        Ball[] balls = new Ball[NUMBER_OF_USERS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
+            System.out.println(balls[i].toString());
+        }
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
+        return Colors.values()[index].toString();
+
     }
 }
--- a/src/main/java/core/basesyntax/ColorSupplier.java
+++ b/src/main/java/core/basesyntax/ColorSupplier.java
@@ -3,9 +3,10 @@ package core.basesyntax;
 import java.util.Random;
 
 public class ColorSupplier {
-    public String getRandomColor() {
-        int index = new Random().nextInt(Colors.values().length);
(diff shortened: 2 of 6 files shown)


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball firstBall = new Lottery().getRandomBall();
```
Elena-Bruyako: [Use for loop for creating several objects of the same class](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-for-loop-for-creating-several-objects-of-the-same-class)

Elena-Bruyako: [Don't create several instances of the same class if you can use only one instance for your purposes](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#dont-create-several-instances-of-the-same-class-if-you-can-use-only-one-instance-for-your-purposes)

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
     public String getRandomColor() {
-        return null;
+        int index = new Random().nextInt(Colors.values().length);
```
Elena-Bruyako: Random should be class level variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        return Colors.values()[index].toString();
```
Elena-Bruyako: [Use name() for getting String representation of enum constants](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#use-name-for-getting-string-representation-of-enum-constants)

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,9 @@
+package core.basesyntax;
+
+public enum Colors {
```
Elena-Bruyako: according to the naming convention enum should be singular
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    public Ball getRandomBall() {
+        ColorSupplier colorSupplier = new ColorSupplier();
```
Elena-Bruyako: ColorSupplier and Random should be class-level variables

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,12 @@
...
+
+        return new Ball(colorSupplier.getRandomColor(),
+                new Random().nextInt(101));
```
Elena-Bruyako: [All magic numbers in your code should be constants.](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#all-magic-numbers-in-your-code-should-be-constants)

Elena-Bruyako (changes requested):
Always pay attention to the [common mistakes file](https://github.com/mate-academy/jv-lottery/blob/master/checklist.md#common-mistakes-jv-lottery)

src/main/java/core/basesyntax/ColorSupplier.java
anurmetov: could u explain why? Thanks!

src/main/java/core/basesyntax/Lottery.java
anurmetov: could u explain purpose of doing it and when should i declare variables as class-level variables? Thanks!

src/main/java/core/basesyntax/Lottery.java
anurmetov: Should it be made to edit magic numbers in the future faster if we have a lot of them implemented in the class? For example : one time changed - all variables in the class are changed as well?

src/main/java/core/basesyntax/ColorSupplier.java
Elena-Bruyako: you don't need to create Random object every time you call getRandomColor method

src/main/java/core/basesyntax/Lottery.java
Elena-Bruyako: answered above

src/main/java/core/basesyntax/Lottery.java
Elena-Bruyako: Correct, try to avoid hardcoding and use constants for numbers, literals, symbols, etc

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
...
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
+
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Ball[] balls = new Ball[NUMBER_OF_USERS];
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = new Lottery().getRandomBall();
+            System.out.println(balls[i].toString());
+        }
```
Elena-Bruyako:
```suggestion
        Lottery lottery = new Lottery();
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,14 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
```
Elena-Bruyako:
```suggestion
    private static final int NUMBER_OF_BALLS = 3;
```

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
+package core.basesyntax;
+
+import java.util.Random;
+
+public class Lottery {
+    private static final int limitOfBallNumber = 101;
```
Elena-Bruyako: create name according to the constant naming convention

src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,14 @@
...
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        return new Ball(colorSupplier.getRandomColor(),
```
Elena-Bruyako: it's better to create local variables for random color and number

Elena-Bruyako (changes requested):

liliia-ponomarenko (approved):
Good job!

PR #1900, opened 2025-08-22
Feature/completion of the lottery task

DianaKiyan:


src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
 package core.basesyntax;
 
+import java.util.Random;
+
 public class ColorSupplier {
-    public String getRandomColor() {
-        return null;
+    final String getRandomColor() {
```
Elena-Bruyako, line 6:
```suggestion
    public String getRandomColor() {
```

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        Random random = new Random();
```
Elena-Bruyako, line 7: Random should be class-level variable

src/main/java/core/basesyntax/ColorSupplier.java
```diff
@@ -1,7 +1,11 @@
...
+        int ballId = random.nextInt(Colors.values().length);
+        return Colors.values()[ballId].toString();
```
Elena-Bruyako, line 9: use `name()` for enum instead of `toString() `
read common mistakes file in the task description

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
+package core.basesyntax;
+
+public enum Colors {
```
Elena-Bruyako, line 3: enum name should be plural according to the naming convention
```suggestion
public enum Color {
```

src/main/java/core/basesyntax/Colors.java
```diff
@@ -0,0 +1,5 @@
...
+    Black, White, Green, Pink, Yellow, Blue, Orange
```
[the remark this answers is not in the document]
Elena-Bruyako: fix comment (apply for other values)
```suggestion
    BLACK, WHITE, Green, Pink, Yellow, Blue, Orange
```

Elena-Bruyako (changes requested):
Good job! Let's improve your solution

PR #1928, opened 2025-10-02
Implement lottery task solution

Inna1989F:


Inna1989F (commented):
Can you look on changing please

PR #2048, opened 2026-04-29
the task is done (I wanna die)

Tacha-hub:


src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int NUMBER_OF_USERS = 3;
```
fetis-off:
```suggestion
    private static final int NUMBER_OF_BALLS = 3;
```
Why variable called NUMBER_OF_USERS? Do we create users?)
Please rename it

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,16 @@
...
+
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        Ball[] balls = new Ball[NUMBER_OF_USERS];
+
+        for (int i = 0; i < balls.length; i++) {
+            balls[i] = lottery.getRandomBall();
+            System.out.println(balls[i]);
```
fetis-off:
```suggestion
            System.out.println(lottery.getRandomBall());
```
Do not create redundant variables

fetis-off (changes requested):
Good work. Left minor comments

src/main/java/core/basesyntax/Application.java
```diff
@@ -1,7 +1,15 @@
 package core.basesyntax;
 
 public class Application {
+    private static final int AMOUNT_OF_USERS = 3;
```
Elena-Bruyako: why is constant name amount of users?

src/main/java/core/basesyntax/Ball.java
```diff
@@ -0,0 +1,29 @@
...
+    public void setRandomNumber(int number) {
+        this.number = number;
+    }
+
+    @Override
+    public String toString() {
+        return "Color is" + " " + color.name() + " " + "number is " + number;
+    }
+
+    public enum Color {
```
Elena-Bruyako: make it as separate class

Elena-Bruyako (changes requested):
see comments

src/main/java/core/basesyntax/Application.java
Tacha-hub: > why is constant name amount of users?

Because with all task I am confused about where, what, and in which task I need to fix. Thank you

fetis-off (approved):
Good one!

PR #2067, opened 2026-06-21
Alpha 1.0

drpacci:


src/main/java/core/basesyntax/Application.java
```diff
@@ -3,5 +3,11 @@
 public class Application {
     public static void main(String[] args) {
         // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako, line 8: 3 is a magic number, could be constant

Elena-Bruyako (approved):
Good job
see comment for the future

PR #2090, opened 2026-08-01
implement ColorSupplier and Lottery with random Ball generator

goffa-exe:


src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
+            System.out.println(lottery.getRandomBall().toString());
```
Elena-Bruyako:
```suggestion
            System.out.println(lottery.getRandomBall());
```

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,10 @@
 
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+
+        for (int i = 0; i < 3; i++) {
```
Elena-Bruyako: 3 is a magic number, create constant

Elena-Bruyako (changes requested):
see comments

src/main/java/core/basesyntax/Application.java
goffa-exe: not sure I understood you correctly

src/main/java/core/basesyntax/Application.java
```diff
@@ -2,6 +2,11 @@
...
 public class Application {
     public static void main(String[] args) {
-        // create three balls using class Lottery and print information about them in console
+        Lottery lottery = new Lottery();
+        final int magicNumber = 3;
+
+        for (int i = 0; i < magicNumber; i++) {
+            System.out.println(lottery.getRandomBall());
+        }
     }
```
Elena-Bruyako, line 11:
```suggestion
        private static final int NUMBER_OF_BALLS = 3;

    public static void main(String[] args) {
        Lottery lottery = new Lottery();
        for (int i = 0; i < NUMBER_OF_BALLS; i++) {
            System.out.println(lottery.getRandomBall());
        }
    }
}
```

Elena-Bruyako (approved):
pay attention to the comment for the future

PR #2095, opened 2026-08-08
jv-lottery-ds1

zormad9:


src/main/java/core/basesyntax/Lottery.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Random;
+
+public class Lottery {
+    private static final int MAX_NUMBER = 100; 
+    private final ColorSupplier colorSupplier = new ColorSupplier();
+    private final Random random = new Random();
+
+    public Ball getRandomBall() {
+        Color randomColor = colorSupplier.getRandomColor();
+        int randomNumber = random.nextInt(MAX_NUMBER) + 1;
```
mateuszwojtkowiak, line 12: 1 is a magic number. It would be better to create constant:

`private static final int MAX_BOUND = 101;
`

mateuszwojtkowiak (approved):
Great job!

